PR #41, opened 2015-04-17
495

elvisciotti:


elvisciotti:
session expire into paramters.yml
session-only login, cookie_lifetime=0, using eventListener
expires or session wiped add a message to login page
unit test
login goes to last accessed url in case of timeout or session wiped

PR #142, opened 2015-07-07
content changes

abisola:
changed firstname to first name and lastname to last name


src/AppBundle/Resources/translations/validators.en.yml
```diff
@@ -36,10 +36,10 @@ user:
     notBlank: The main phone number cannot be blank
 client:
     firstname:
-        notBlank: Please enter your firstname
+        notBlank: Please enter your first name
         minMessage: Firstname must be at least {{ limit }} characters
     lastname:
-        notBlank: Please enter your lastname
+        notBlank: Please enter your last name
         minMessage: Lastname must be at least {{ limit }} characters
```
elvisciotti: @abisola "Lastname" does this  need a rename too ?

PR #202, opened 2015-07-31
356

sophycolbert:
Update the accounts text on the overview page

### DIFF
--- a/src/AppBundle/Resources/translations/report-overview.en.yml
+++ b/src/AppBundle/Resources/translations/report-overview.en.yml
@@ -34,14 +34,14 @@ pages:
         subSectionList: "*buying or selling or renting property *buying or selling investments or shares *loans *making gifts *moving the client to another nursing or care home"
       accounts:
         subSectionTitle: "Accounts"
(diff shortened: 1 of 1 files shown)


src/AppBundle/Resources/translations/report-overview.en.yml
```diff
@@ -34,14 +34,14 @@ pages:
         subSectionList: "*buying or selling or renting property *buying or selling investments or shares *loans *making gifts *moving the client to another nursing or care home"
       accounts:
         subSectionTitle: "Accounts"
-        subSectionDescription: "Add details of your client's accounts. Let us know the regular payments you make and money you receive for the client."
+        subSectionDescription: "Add details of your client’s accounts. We need to know the totals for the different types of payments you make and money you receive for the client. It's easiest to fill this in toward the end of your reporting period, when you know the final total amounts."
```
elvisciotti: @sophycolbert  when on staging, check if this is rendered OK. It will go on more than line, and the apostrophe sign could be rendered wrongly

PR #216, opened 2015-08-14
Warn user if leaving transactions page without saving

abisola:
This is only for the transactions page in response to current user issues

### DIFF
--- a/src/AppBundle/Resources/views/Account/account.html.twig
+++ b/src/AppBundle/Resources/views/Account/account.html.twig
@@ -115,7 +115,7 @@ activeTab: "tab-accounts"
 
     <script type="text/javascript">
         var transactionHasChanged = false;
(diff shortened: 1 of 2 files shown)


ztolley:
Can you let me know when you have done some cross browser testing on this before we merge it

src/AppBundle/Resources/views/Account/account.html.twig
```diff
@@ -113,6 +113,21 @@ activeTab: "tab-accounts"
 
         {{ form_end(form) }}
 
+    <script type="text/javascript">
+        var transactionHasChanged = false;
+        var message = "You haven't saved the transactions you entered. Do you want to leave without saving?";
```
elvisciotti: that could stay in the translation file (backslashing double quotes)

var message = '{{ 'file.key.subkey ' | trans | escape('js') }}'

PR #227, opened 2015-08-26
fix: known date legend

elvisciotti:


src/AppBundle/Resources/views/Components/Form/_known-date.html.twig
```diff
@@ -2,7 +2,7 @@
 <div class="form-group form-date {% if not element.vars.valid %}field-with-errors behat-region-form-errors{% endif %}" id="{{ id }}">
     <fieldset>
         <legend class="form-label">
-            {% if legendTextJS is defined %}
+            {% if legendTextJS %}
```
elvisciotti: is always defined, but can be null, so this is the needed check

PR #251, opened 2015-09-03
content-03-09-15

sophycolbert:
Amend page titles for consistency/house style
Remove lorem text and replace with “ “


elvisciotti:
@sophycolbert could you please remove the lines that used to contain the "Lorem ipsum" text instead of setting them to empty string ?
when done I'll merge straighaway

elvisciotti:
@sophycolbert  I've removed the subtitle lines myself, and merged

PR #282, opened 2015-10-02
Log location

munhitsu:


app/config/config.yml
```diff
@@ -106,3 +106,17 @@ parameters:
         14: assets_held_outside_england_and_wales
         15: court_funds_office_accounts 
         16: other_assets
+
+services:
+    logstash_formatter:
+        class: Monolog\Formatter\LogstashFormatter
+        arguments:
+        - application.client
```
elvisciotti: not sure it needs an additional tab here

PR #352, opened 2015-12-14
display opg_docker_tag in footer

munhitsu:


ztolley:
what does it render like?

PR #364, opened 2015-12-31
accounts_mk2-1296-balance-extra-info css and format

elvisciotti:


src/AppBundle/Resources/assets/scss/digideps/_balance-summary-table.scss
```diff
@@ -117,3 +117,15 @@ table.money-table {
     }
     
 }
+
+
+.balance-warning.page-section-warning {
```
elvisciotti: @ztolley this is to align the "warning" text right to the icon, in compliance with the design in the PDF. I didn't want to apply to all the "page-section-warning" classes in order to mess other things up, so I added a "balance-warning" class to differentiate.

PR #388, opened 2016-02-04
Dd 1363

ztolley:
Fixed autosave issues using Delete key, clear on Windows 10 and removing fields with descriptions.


src/AppBundle/Resources/views/Layouts/application.html.twig
```diff
@@ -120,7 +120,7 @@
 
 
 {% block javascripts %}
-    <script type="text/javascript" src="{{ 'javascripts/application.min.js'| assetUrl }}"></script>
+    <script type="text/javascript" src="{{ 'javascripts/application.js'| assetUrl }}"></script>
```
elvisciotti: @ztolley

PR #411, opened 2016-04-05
DD-1417

sophycolbert:
Update question about other charges on the property.


sophycolbert:
Hi Elvis - I added a 'hint' line - is this OK?

PR #452, opened 2016-09-14
progress_bar_refactor

elvisciotti:
…c into extension and out of yml


elvisciotti:
@Biggsen I've changed the progress_bar helper to support ODR "toggling". I'll merge master into ODR

PR #459, opened 2016-10-05
DDPB-412-transfers-to-other-accounts

lauratennant:
Have clarified wording


elvisciotti:
@lauratennant I have to create the backend to support %client%, it's not automatic for all the translations at the moment

PR #467, opened 2016-10-13
DDPB 412 transfers to other accounts

lauratennant:
Shortened text (because it’s not currently wrapping


elvisciotti:
@lauratennant merged

PR #483, opened 2016-12-20
Ddpb 450 ndr due date

sheanie:


src/AppBundle/Resources/views/Admin/_edit.html.twig
```diff
@@ -31,6 +31,12 @@ transDomain: 'admin'
         <li>Client name: {{ firstClient.firstname }}</li>
         <li>Client lastname: {{ firstClient.lastname }}</li>
         <li>Case number: {{ firstClient.caseNumber }}</li>
+        {% if firstClient.odr %}
```
elvisciotti: @sheanie very minor thing. You can use "and" with twig if you want to collapse the two nested "if"

src/AppBundle/Resources/views/Odr/Index/_report_row.html.twig
```diff
@@ -20,10 +20,7 @@
                 {% endif %}
             </p>
             <p class="text-secondary soft-half--top">
-                {% if reportActive %}
```
elvisciotti: @sheanie reportActive (as far as I remember) was the current report for clients (among the potential already-submitted ones). Is it not needed anymore with this feature ?

elvisciotti:
@sheanie  happy to merge. The only bug since last time I saw it was the existing data not having ODR (or something similar), or there is something more I missed ?

PR #484, opened 2016-12-21
OPGOPS-1677 Inital dev setup of digideps

carinadigital:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.service.opg.digital/opguk/php-fpm
+FROM registry.service.opg.digital/opguk/php-fpm:0.1.207
```
brettminnie: Why are we locking to a version here? I was under the impression we removed this so that when built latest stable would always be pulled in to make sure security updates etc where applied

elvisciotti: @brettminnie  I suggested that after we had a disruption due to a non-working version. If there is an update on php-fpm I prefer to change the Dockerfile in a branch and test it there, unless you guarantee php-fpm latest version is always stable

carinadigital: I think this depends on whether its worse to forget to update or to have some instability.

brettminnie: Fair do, lets merge then have that discussion again

PR #498, opened 2017-01-12
status-service-test complete and refactor ReportStatusService

elvisciotti:


elvisciotti:
@sheanie FYI. Some tests were missing from the report validation service used to understand when the report is ready for submissions. I needed that class to be complete for 103. All passing (build #1​51​​0)

PR #508, opened 2017-01-19
DDPB-765-decisions-column-heading-yes-no

lauratennant:
@biggs Column heading 3 amended from:
How was client involved
to:
If not, explain why; if yes, provide detail


Biggsen:
Not pretty, but still an improvement. Approved.

PR #514, opened 2017-01-23
DDPB-860-check-through-amends-to-NDR

lauratennant:
@biggs Have rewritten help text on the homepage for your reports. Think it
makes more sense now.


elvisciotti:
@lauratennant merged

PR #519, opened 2017-01-24
Ddpb 689 gds details

Biggsen:
custom accordian replaced with native <details>


Gulpfile.js
```diff
@@ -187,7 +189,7 @@ gulp.task('lint.sass', function() {
 
 // Check javascript follows some good guidelines and check for obvious errors.
 gulp.task('lint.js', function () {
-    gulp.src(config.jsSrc + '/**/*.js')
+    gulp.src([config.jsSrc + '/**/*.js','!'+ config.jsSrc + '/**/details.polyfill.js'])
```
elvisciotti: what's the semantic of "!" ?

src/AppBundle/Resources/assets/javascripts/modules/accounttransactions.js
```diff
@@ -1,3 +1,6 @@
+/* globals jQuery */
```
elvisciotti: I think this module can be deleted

Gulpfile.js
Biggsen: exclude

src/AppBundle/Resources/assets/javascripts/modules/accounttransactions.js
Biggsen: all of JS needs looking at. probably loads we can throw out.

PR #544, opened 2017-02-07
DDPB-929 modernising head link refs, esp ie conditionals.

Biggsen:
IE8 radio buttons and checkboxes were failing because application.css was being loaded as well as application-ie8.css. I updated all the link refs to match a more modern version of gov.uk. This fixes the loading issue.

Also removed a legacy font loader:
https://github.com/alphagov/govuk_template/pull/219

Shifted the stylesheets into moj_template. No need to keep them in application.

Tested in a feature branch, looks good.

### DIFF
--- a/src/AppBundle/Resources/views/Layouts/application.html.twig
+++ b/src/AppBundle/Resources/views/Layouts/application.html.twig
@@ -8,29 +8,11 @@
 
 {% block topOfPage %}{% endblock %}
 
-{% block stylesheets %}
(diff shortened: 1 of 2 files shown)


elvisciotti:
@Biggsen MOJ twig template - as far as I remember was generated from another given MOJ template. I saw your PR on the other repo so I suppose your changes are aware of that

bodyClasses removal <-- well spotted, not used anymore

PR #550, opened 2017-02-12
DDPB-917 notifications component refactored into a macro

Biggsen:


src/AppBundle/Resources/views/Macros/macros.html.twig
```diff
@@ -120,4 +120,40 @@
         'labelText': 'addAnotherPage.save' | trans({}, 'common'),
         'buttonClass': 'behat-link-save-and-continue',
     }) }}
+{% endmacro %}
+
+{% macro notification(alertType,message) %}
+    {% if alertType is defined %}
+        {% if alertType == 'notice' %}
+            <div class="alert-notice panel panel-border-narrow">
```
elvisciotti: alternative approach would be use the same piece of HTML and use variables to define the CSS classes

PR #561, opened 2017-02-20
Ddpb 975 css cleanout

Biggsen:
Feckloads of css killed!!!


elvisciotti:
can't find any problem with it

PR #567, opened 2017-03-01
Ddpb 1037 get referer safe

elvisciotti:


src/AppBundle/Controller/AbstractController.php
```diff
@@ -172,4 +174,29 @@ protected function stepRedirector()
...
+        try {
+            $routeParams = $this->get('router')->match($refererUrlPath);
+        } catch (ResourceNotFoundException $e) {
+            return null;
+        }
+        $routeName = $routeParams['_route'];
+        if (in_array($routeName, $excludedRoutes)) {
+            return null;
+        }
+        unset($routeParams['_route']);
```
slizzio: Just a question, why is this being unset here?

elvisciotti: _route gets otherwise passed as a param, that doesn't exist in the route itself and gets rendered as a useless param &route=<routeName>

PR #568, opened 2017-03-01
Ddpb 1039 xss client name

elvisciotti:


Biggsen:
looks good to me.
agreed would be good to find a global solution. seems overly redundant. someone put it in the backlog ;)?

elvisciotti:
I was thinking of injecting the client firstname globally, but that variable is only available after the controller does an API call and gets the client. We'd need to place some logic between that and the twig rendering, either by assuming the client name variable is the same or putting some logic in each controller to do that (but that would only move the complication from the twig template into the controller).

Other things to consider
1. in some cases we need the client name, and other variables from the view to pass to replace in the translation key, so the "magic" of having %client% would make this confusing.
2.in some templates, the transOptions are used many times
3. a "magic" approach would make it difficult to debug. e.g. when the report/client variable is available in the controller 

To avoid the repeating of the transOptions array, we could create a twig method for that like transOptions(client), to pass in the twig template when needed

PR #571, opened 2017-03-03
DDPB-992 html/css for admin pages improved

Biggsen:


src/AppBundle/Resources/views/Admin/editUser.html.twig
```diff
@@ -33,53 +38,58 @@
             {{ form_checkbox(form.odrEnabled, 'addUserForm.odrEnabled') }}
         </div>
     {% endif %}
-    {{ form_submit(form.save,'addUserForm.submit', {'buttonClass': 'behat-link-save'}) }}
+    {{ form_submit(form.save,'editUserForm.submit', {'buttonClass': 'behat-link-save'}) }}
 
+    {% if action is defined %}
```
elvisciotti: not sure if this one is still needed

PR #579, opened 2017-03-07
Ddpb 1019

elvisciotti:


src/AppBundle/Entity/User.php
```diff
@@ -693,6 +713,44 @@ public function setAdManaged($adManaged)
...
+    /**
+     * @param string $paTeamName
+     * @return User
+     */
+    public function setPaTeamName($paTeamName)
+    {
+        $this->paTeamName = $paTeamName;
+
+        return $this;
+    }
```
slizzio: To remove getter / setter

src/AppBundle/Form/User/UserDetailsFullType.php
```diff
@@ -31,7 +17,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         ->add('addressPostcode', 'text')
         ->add('addressCountry', 'country', [
             'preferred_choices' => ['', 'GB'],
-            'empty_value' => $this->addressCountryEmptyValue,
+            'empty_value' => 'Please select ...',
```
slizzio: how is this translated? would expect a translation key here

elvisciotti: it was originally injected in the constructor, but it was never edited since the start, so I reduced the complexity and left it in English inside

elvisciotti: had no time to check if empty_value are translated in Symfony after I've upgraded it

src/AppBundle/Entity/User.php
```diff
@@ -232,6 +232,26 @@ class User implements AdvancedUserInterface
...
+
+    /**
+     * @JMS\Type("string")
+     * @JMS\Groups({"user_details_pa"})
+     * @Assert\NotBlank( message="user.paTeamName.notBlank", groups={"user_details_pa"} )
+     * @Assert\Length(max=50, maxMessage="user.paTeamName.maxMessage", groups={"user_details_pa"} )
+     *
+     * @var string
+     */
+    private $paTeamName;
```
elvisciotti: I've removed it from the API User.paTeamName, but this entity is associated to the Form to save/load the team name. We'll rediscuss in the next PR where I'll show the implementation (with logic to store/load team name in the API, and not in the client)

PR #585, opened 2017-03-09
[DO NOT MERGE] DDPB-788 Added Team Controller, AddUser template  and translations

slizzio:


src/AppBundle/Resources/views/Pa/Team/addUser.html.twig
```diff
@@ -0,0 +1,31 @@
...
+{% block breadcrumbs %}
+    {{ macros.breadcrumbsSettings('user-add') }}
+{% endblock %}
+
+{% block pageContent %}
+    {{ form_start(form, {attr: {novalidate: 'novalidate' }}) }}
+
+    {% include 'AppBundle:User:_form_fields.html.twig' %}
+
+    {{ form_submit(form.save,'form.editYourDetails.controls.save', {'buttonClass': 'behat-link-save'}) }}
```
slizzio: not sure about using editYourDetails. However im unsure where/how these controls are set?

src/AppBundle/Controller/Pa/TeamController.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+class TeamController extends AbstractController
+{
+    /**
+     * @Route("/add-user", name="pa_team_user")
+     * @Template
+     */
+    public function addUserAction(Request $request)
+    {
+        $form = $this->createForm(new FormDir\User\UserDetailsPaType());
```
elvisciotti: the radio box "Grant user administrator access" is needed in this page. What do yo think about another form inheriting this one ?

src/AppBundle/Controller/Pa/TeamController.php
```diff
@@ -0,0 +1,42 @@
...
+
+        $form->handleRequest($request);
+
+        if ($form->isValid()) {
+            $formData = $form->getData();
+
+            $this->getRestClient()->post('user/add', $formData);
```
elvisciotti: as you can see from the other "post" calls, it's normally a good practice to specify the JMS group you want to pass to the API. In some cases JMS might send more than expected, and either having the API failing for timeout, or saving data you didn't want to

src/AppBundle/Resources/translations/user-account.en.yml
```diff
@@ -1,8 +1,9 @@
 h1Title:
     client-edit: "Account: %client%'s details"
     deputy-edit: "Account: Your details"
-    change-password: "Account: Change password"    
+    change-password: "Account: Change password"
+    add-account: "Add a user account"
```
elvisciotti: @Biggsen to decide if a different template is needed here. I'd suggest so, the user-account yml is used in many other views and might confuse Content editor

src/AppBundle/Resources/views/Pa/Team/addUser.html.twig
elvisciotti: 2nd param is the path to the translation file. See FormFieldsExtension::renderFormSubmit().
If you use a new translation file, you can decide what's the path. e.g. form.save.submit, and have this in the new YML file

form:
  save:
  label: Add user

PR #596, opened 2017-03-13
DDPB Added ACL

slizzio:


elvisciotti:
@slizzio  code is correct. You can some of the ACL from symfony security config and protect URLs. In this case we have different roles so you can easily define rules here https://github.com/ministryofjustice/opg-digi-deps-client/blob/master/app/config/security.yml#L63

elvisciotti:
@slizzio also, grep for "isGranted" in the views and see how ACL is already implemented in those files. e.g. main template navigation bar

src/AppBundle/Controller/Pa/TeamController.php
```diff
@@ -62,6 +63,10 @@ public function editAction(Request $request, $id)
     {
         $user = $this->getRestClient()->get('team/member/'.$id, 'User');
 
+        if (!$this->getUser()->canEditUser($user)) {
```
elvisciotti: that could be a security.yml rule. The API also should be implemented not to give access to edit users depending on the roles

src/AppBundle/Entity/User.php
```diff
@@ -799,4 +799,60 @@ public function isDeputy()
...
+    }
+
+    /**
+     * Is User the Named Deputy?
+     *
+     * @return bool
+     */
+    private function isNamedDeputy()
+    {
+        return $this->getRoleName() === User::ROLE_PA;    }
```
elvisciotti: formatting here and before

app/config/security.yml
```diff
@@ -60,7 +60,9 @@ security:
         # /admin accessible only from ADMIN, super users and AD users
         - { path: ^/admin, roles: [ROLE_ADMIN, ROLE_AD] }
         - { path: ^/ad, roles: [ROLE_AD] }
-        - { path: ^/pa, roles: [ROLE_PA, ROLE_PA_ADMIN, ROLE_PA_TEAM_MEMBER] }
+        - { path: /pa/$, roles: [ROLE_PA, ROLE_PA_ADMIN, ROLE_PA_TEAM_MEMBER] }
```
elvisciotti: check that this doesn't conflict with filters (search, paginations, tabs)

elvisciotti: http://symfony.com/doc/current/security/access_control.html
" the first rule that matches is used"
So, I think you could keep /pa last to be accessible from any PA, and move your more strictly rules around edit/add/delete before that

src/AppBundle/Controller/Pa/TeamController.php
elvisciotti: this check should be also done in the API, that throws exception. So, this could be done by the API throwing access denied exception, caught by the client and re-diplayed, and the logic basically moved from client into the API

app/config/security.yml
slizzio: Done

src/AppBundle/Entity/User.php
slizzio: Done

PR #598, opened 2017-03-15
DDPB-1074 content tweaks for 102

Biggsen:


src/AppBundle/Resources/translations/odr-visits-care.en.yml
```diff
@@ -20,11 +20,13 @@ summaryPage:
     weAskAbout: In this section, we ask you about %client%'s care and contact with you and other people.
 
 form:
-  doYouLiveWithClient.label: Do you live with %client%?
+  doYouLiveWithClient:
+    label: Do you live with %client%?
   howOftenDoYouContactClient:
     label: |
```
elvisciotti: @Biggsen is the old label still used, after labelShort has been added ?

Biggsen: Yes. It's still used in the form. The short one is used in the answer table.

PR #599, opened 2017-03-17
DDPB-1031 Add user refined

Biggsen:


src/AppBundle/Resources/views/Pa/Team/add.html.twig
```diff
@@ -7,13 +7,10 @@
...
 {% block pageTitle %}{{ 'addPage.h1Title' | trans }}{% endblock %}
-
-{% block linkBack %}
-    <a href="{{ path('pa_team') }}" class="link-back behat-link-back">Back</a>
-{% endblock %}
+    {% block pageTitleClass %}heading-large{% endblock %}
 
 {% block breadcrumbs %}
-    {{ macros.breadcrumbsSettings('user-add') }}
+    [breadcrumbs]
```
elvisciotti: is that wanted ?

Biggsen: Those breadcrumbs were not working right, so I put in a placeholder [breadcrumbs] instead. To answer your question, no, that is not wanted. We'll have to have a breadcrumbs ticket and do them all properly.

PR #601, opened 2017-03-17
pa-voters-acl working version of voters

elvisciotti:


elvisciotti:
@slizzio I've implemented an ACL using Symfony Voters 
http://symfony.com/doc/2.8/security/voters.html

The same logic is used in controller actions and views with a simple line. Logic is in the same place.
This is much better than the conditional logic in the controllers. All this work OK after my tests. Merge if tests is passing as it's already a better improvement.

PR #606, opened 2017-03-27
OPGOPS-1836: compile assets during instance startup

Azulinho:
From: https://opgtransform.atlassian.net/browse/OPGOPS-1836

This PR makes sure we compile the assets during docker instance startup, allowing us to use a local volume with uncommited changes but also supporting (later on) baking all the assets directly into the docker image.


docker/my_init.d/81-recompile-assets
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# recompile assets at docker instance boot time
```
carinadigital: If you look at opg-core-docker, we can use the makefile to compile the assets without installing the build tools in the container.

Azulinho: I saw that, the reason I added that step was to be able to have a single docker image with all the assets baked in and not rely on an external mounted volume

Azulinho:
dropping this as : ministryofjustice/Bugfix-gulp-clean-dependecy fixes the pain I was having here

PR #609, opened 2017-03-29
‘clean’ task is now a dependency for any tasks saving into web/assets

Biggsen:


Gulpfile.js
```diff
@@ -136,7 +137,7 @@ gulp.task('images', () => {
 
 // Creates the production version of the service javascript.
 // Files are concatinated and then minified with uglify.
-gulp.task('js.prod', ['lint.js'], () => {
+gulp.task('js.prod', ['lint.js', 'clean'], () => {
```
carinadigital: `function` before () ?

PR #618, opened 2017-04-05
Ddpb 1096

slizzio:


Biggsen:
Made a few tweaks to the additional validation messages (minMessage & maxMessage). Replaced a few 'Your' with 'The' so it could be used in both cases.

PR #633, opened 2017-04-24
DDPB 1149

iterative-it:


tests/behat/sql/pa/pa-teardown.sql
```diff
@@ -0,0 +1,4 @@
+--Clear all data and users required by pa team tests
+truncate table report,deputy_case,client,user_team,dd_team RESTART IDENTITY cascade;
+DELETE FROM dd_user WHERE id > 4;
```
slizzio: can we not delete users based on roles? rather than IDs

tests/behat/sql/pa/pateam-setup.sql
```diff
@@ -0,0 +1,60 @@
+--Sets up the data required to run the pa team tests.
+--Requires a clean database
+INSERT INTO dd_team (team_name) VALUES (NULL);
+INSERT INTO dd_user (firstname, lastname, password, email, active, registration_date, registration_token, token_date, address_postcode, address_country, phone_main, last_logged_in, deputy_no, odr_enabled, ad_managed, role_name, job_title, agree_terms_use, agree_terms_use_date)
+VALUES ('John Named', 'Green', '9k4PZrYAhWIMcVCELlGk/xJmzYtFLGmta924lBP/VvM4T7sfEDomfn373dueeyh+CADl/aPlzOQV0h+3h1N3Wg==', 'behat-pa1@publicguardian.gsi.gov.uk', TRUE, CURRENT_TIMESTAMP, '', CURRENT_TIMESTAMP, 'SW1', 'GB', '10000000001', CURRENT_TIMESTAMP, 9000001, NULL, NULL, 'ROLE_PA', 'Solicitor', true, CURRENT_TIMESTAMP);
+INSERT INTO user_team VALUES (5, 1);
```
slizzio: Same here... it might be the only way, but relying on ids is a bit risky. as changes to other scripts may cause these to break or worse, set the database up in an unexpected state.

PR #635, opened 2017-04-26
DDPB-1172: Fixing encoding on PDF generation

iterative-it:
This was an encoding problem when creating the PDF. I have fixed this and pound signs and special characters now render correctly

### DIFF
--- a/src/AppBundle/Service/WkHtmlToPdfGenerator.php
+++ b/src/AppBundle/Service/WkHtmlToPdfGenerator.php
@@ -54,11 +54,13 @@ class WkHtmlToPdfGenerator
 
         $ch = curl_init();
 
+        curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, $this->timeoutSeconds);
+        curl_setopt($ch, CURLOPT_TIMEOUT, $this->timeoutSeconds); //timeout in seconds
         curl_setopt($ch, CURLOPT_URL, $this->url);
         curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'POST');
         curl_setopt($ch, CURLOPT_POSTFIELDS, $dataString);
         curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/WkHtmlToPdfGenerator.php
```diff
@@ -39,17 +39,30 @@ public function isAlive()
...
+        );
+        $dataString = json_encode($data);
+        $headers = array(
+            'Content-Type: application/json',
+            'Content-Length: ' . strlen($dataString),
+        );
+
         $ch = curl_init();
-        curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, $this->timeoutSeconds);
-        curl_setopt($ch, CURLOPT_TIMEOUT, $this->timeoutSeconds); //timeout in seconds
```
slizzio: How are timeouts handled?

src/AppBundle/Service/WkHtmlToPdfGenerator.php
```diff
@@ -39,17 +39,30 @@ public function isAlive()
...
-        curl_setopt($ch, CURLOPT_HTTPHEADER, ['Content-type: application/json']); // Assuming you're requesting JSON
-        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
-        $body = json_encode([
-            'contents' => base64_encode($html),
-        ]);
-        curl_setopt($ch, CURLOPT_POSTFIELDS, $body);
+        curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'POST');
+        curl_setopt($ch, CURLOPT_POSTFIELDS, $dataString);
+        curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
```
slizzio: are you sure the curl library on production matches your local curl library? May require 1 rather than true

PR #648, opened 2017-05-09
DDPB-1145: Behat tests to verify links in header for lay and pa

iterative-it:
Seems that the links were already in place so the only change was to add behat tests to verify

### DIFF
--- a/tests/behat/features/deputy/01-registration-steps/04-add-client-and-report.feature
+++ b/tests/behat/features/deputy/01-registration-steps/04-add-client-and-report.feature
@@ -172,3 +172,14 @@ Feature: deputy / user / add client and report
     Given I click on "reports, report-2016-open"
     Then the URL should match "report/\d+/overview"
     And I save the page as "deputy-report-overview"
+
+  @deputy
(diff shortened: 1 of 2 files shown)


tests/behat/features/deputy/01-registration-steps/04-add-client-and-report.feature
```diff
@@ -172,3 +172,14 @@ Feature: deputy / user / add client and report
...
+
+  @deputy
+  Scenario: report-overview links
+    Given I am logged in as "behat-user@publicguardian.gsi.gov.uk" with password "Abcd1234"
+    #Lay deputy links
+    Then I should see the "user-account" link
+    Then I should see the "reports" link
+    Then I should see the "logout" link
+    #PA links
+    Then I should not see the "pa-dashboard" link
```
elvisciotti: I'd reword a bit the "Then" prefixes
e.g.

Then I should see the "user-account" link
And I should see the "reports" link
And I should see the "logout" link
But I should not see the "pa-dashboard" link

PR #650, opened 2017-05-10
DDPB-1166 update message as per Liz's comment

elvisciotti:


iterative-it:
Would it make sense to add the helpline number to the message? I guess the message itself has come from Liz so it's fine.

PR #653, opened 2017-05-12
DDPB-1209: Attach submitted report pdf to PA email

iterative-it:
PA submission email now has the PDF attached. I've also reworded the email text to reflect this.

Tested and this works correctly. Email received containing the pdf.

All behat tests pass locally.

### DIFF
--- a/src/AppBundle/Resources/views/Email/report-submission-confirm.html.twig
+++ b/src/AppBundle/Resources/views/Email/report-submission-confirm.html.twig
@@ -12,10 +12,14 @@
 
     <p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;">We’re in the process of reviewing your report, and will contact you when this has been done. </p>
 
-     <p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;"><strong>Download a pdf of your report</strong></p>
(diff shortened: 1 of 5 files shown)


src/AppBundle/Resources/views/Email/report-submission-confirm.html.twig
```diff
@@ -12,10 +12,14 @@
 
     <p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;">We’re in the process of reviewing your report, and will contact you when this has been done. </p>
 
-     <p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;"><strong>Download a pdf of your report</strong></p>
-
-    <p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;">To view and download a copy of the report for your records, <a style="word-wrap: break-word;" href="{{ homepageUrl }}">sign back into the deputy report service</a>. Go to the ‘Your reports' homepage, where you’ll find a link to download a pdf of your report.
-    </p>
+    {% if app.user.isDeputyPa() %}
```
elvisciotti: did you test that "app.user" is available when the template is loaded from the MailFactory ? I think so, but never done that so far

PR #656, opened 2017-05-16
DDPB-1111 Added session ini variable settings to further secure sessi…

slizzio:
…on IDs

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -8,6 +8,10 @@ class AppKernel extends Kernel
     public function __construct($environment, $debug)
     {
         date_default_timezone_set('Europe/London');
+        ini_set('session.hash_function', 'sha256');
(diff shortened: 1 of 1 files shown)


app/AppKernel.php
```diff
@@ -8,6 +8,10 @@ class AppKernel extends Kernel
     public function __construct($environment, $debug)
     {
         date_default_timezone_set('Europe/London');
+        ini_set('session.hash_function', 'sha256');
+        ini_set('session.entropy_length', '32');
+        ini_set('session.cookie_httponly', 1);
```
elvisciotti: @slizzio app/config.yml framework.session already contain this option, so - I guess - symfony will override that value when applying the YML config

slizzio: Agreed and removed. Good spot

PR #658, opened 2017-05-17
DDPB-1215 add JMS type for report, in case it gets sent from the API

elvisciotti:


elvisciotti:
those entities dont' normally have the report being sent back from the API. But in case they do, those annotation guarantee it won't fail.

PR #664, opened 2017-05-19
Ddpb 1231 delete account with transfers

slizzio:


iterative-it:
Is it worth doing the error handling for 1231 with an error code rather than assuming the error is due to the transactions still being present?

PR #670, opened 2017-05-25
[DO NOT MERGE]Ddpb 1137 - AD 

elvisciotti:


elvisciotti:
only text changes on AD homepage + very small fix on session variables when deputy side is used by AD + Lay on the same machine (rare, only happens locally)

PR #672, opened 2017-05-30
behat --stop-on-failure flag

elvisciotti:


elvisciotti:
@iterative-it @brettminnie I prefer behat tests to fail at the first failure, or we risk to think they pass locally if we forgot to scroll the screen, and also on jenkins I don't see the point to run all of them after the first failure. If OK I'll merge after the tomorrow's release

brettminnie:
:+1: Thought we had set it up this way, clearly not

iterative-it:
@elvisciotti, @slizzio changed it to be this way as he got tired of fixing one fail, running it again and seeing another fail to fix. Is there any way to achieve this locally but not on the build server?

brettminnie:
Rather it fail fast on the build server as well

elvisciotti:
@iterative-it let's hold this PR for now. Those scripts need to be refactored, adtest.sh doesn't make sense, so I'd prefer to merge them together, and have behat options configurables from outside the script, so we can decide how to run them locally and in the server. I guess there are cases where it's useful to see all the failures in a single run

PR #673, opened 2017-05-30
DDPB-1211 Terms content added. user-terms-use renamed pa-terms-use.

Biggsen:


src/AppBundle/Resources/translations/pa-terms-use.en.yml
```diff
@@ -0,0 +1,54 @@
...
+    For PA deputyships, personal information includes thats gathered about the clients
+    the PA is deputy for, and the user IDs, passwords and email addresses needed to access
+    that information.
+section02:
+  heading: Access to the service
+  para01: |
+    As named deputy you can add a limited number of other colleagues, or users, to help
+    administer the deputyship.
+  para02: |
+    It is then critical that you as public authority named deputy assume sole responsibility for:
```
elvisciotti: @Biggsen is this rendering properly ? I used to have parsing problems when content with semicolons it not wrapped as misintepreted as a key. If it renders, I guess symfony understands that it's text as under "|"

PR #680, opened 2017-06-05
Ddpb 807 view client details

elvisciotti:


src/AppBundle/Entity/Client.php
```diff
@@ -288,13 +303,13 @@ public function getFullname()
 
     public function setCourtDateWithoutTime($courtDate = null)
     {
-        $this->courtDate = ($courtDate instanceof \DateTime) ?
-                new \DateTime($courtDate->format('Y-m-d')) : null;
+        $this->courtDate = ($courtDate instanceof DateTime) ?
+                new DateTime($courtDate->format('Y-m-d')) : null;
```
slizzio: Is it possible to call format on the existing object? rather than creating new instance of DateTime?

src/AppBundle/Entity/Client.php
```diff
@@ -288,13 +303,13 @@ public function getFullname()
...
     }
 
     public function isValidCourtDate(ExecutionContextInterface $context)
     {
-        $today = new \DateTime();
+        $today = new DateTime();
```
slizzio: Can you check this is the correct object? I always used \DateTime but could be wrong

elvisciotti: @slizzio I always did the same, but in this file DateTime was imported and used without the backlash, so I just made the file consistent

src/AppBundle/Entity/Client.php
elvisciotti: @slizzio I didn't initially write this, but I think it's correct. This method wants to assign a date without the time starting from the given date. An alternative could be cloning + setTime(0,0,0) http://php.net/manual/en/datetime.settime.php

src/AppBundle/Entity/Client.php
elvisciotti: @slizzio @iterative-it  I think we need to agree on what to do with fully qualified namespaces vs imported ones for PHP native classes. Most of the code is  using "\DateTime" without importing the namespace in other entities ATM

PR #697, opened 2017-06-26
Ddpb 1311 remove pdf from emails

slizzio:


src/AppBundle/Service/Mailer/MailFactory.php
```diff
@@ -184,17 +184,13 @@ public function createReportEmail(EntityDir\User $user, EntityDir\Report\Report
     {
         $email = new ModelDir\Email();
 
-        $viewParams = [
```
elvisciotti: createReportEmail first param could be removed

src/AppBundle/Service/Mailer/MailFactory.php
```diff
@@ -311,8 +307,7 @@ public function createReportSubmissionConfirmationEmail(EntityDir\User $user, En
      */
     public function createPaReportSubmissionConfirmationEmail(EntityDir\User $user, EntityDir\Report\Report $submittedReport, EntityDir\Report\Report $newReport, $pdfBinaryContent)
```
elvisciotti: This method is now only calling another method with the same parameters. If it's wanted (Pa=Lay, not sure), I'd replace the usage of createPaReportSubmissionConfirmationEmail with createReportSubmissionConfirmationEmail and remove a method overall

src/AppBundle/Service/Mailer/MailFactory.php
slizzio: Removed, UT updated

PR #698, opened 2017-06-27
Ddpb 1309 report review back link

slizzio:


src/AppBundle/Controller/AbstractController.php
```diff
@@ -204,4 +204,33 @@ protected function getRefererUrlSafe(Request $request, array $excludedRoutes = [
...
+            // generate link
+            return $this->generateUrl('report_overview', ['reportId' => $report->getId()]);
+        }
+
+        $this->get('logger')->log(
+            'warning',
+            'Client entity missing current report when trying to generate client profile link'
+        );
+
+        throw new \Exception('Unable to generate client profile link.');
```
elvisciotti: exception message should already create a log entry in kibana

PR #699, opened 2017-06-27
Ddpb 1333 - edit report dates from PA

elvisciotti:


src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -52,50 +52,61 @@ class ReportController extends AbstractController
...
-            }
+    /**
+     * Edit single report
+     *
+     * @Route("/reports/edit/{reportId}", name="report_edit")
+     * @Template()
+     */
+    public function editAction(Request $request, $reportId)
+    {
+        $report = $this->getReport($reportId);
```
slizzio: should this be getReportIfNotSubmitted() ?

PR #717, opened 2017-07-19
DDPB-1175 pa password

jackgoodby:


src/AppBundle/Resources/views/Pa/Settings/profilePasswordEdit.html.twig
```diff
@@ -0,0 +1,29 @@
+{% extends 'AppBundle:Layouts:application.html.twig' %}
+
+{% import 'AppBundle:Macros:macros.html.twig' as macros %}
+
+{% set translationDomain = "pa-profile" %}
+{% trans_default_domain translationDomain %}
+
+{% block htmlTitle %}Settings - change password{% endblock %}
```
elvisciotti: it'd be nice to have those into the translation file, more aligned with the rest of PA templates

src/AppBundle/Controller/Pa/PaSettingsController.php
```diff
@@ -81,4 +81,33 @@ public function profileEditAction(Request $request)
...
         ];
     }
+
+    /**
+     * Display the password edit page
+     *
+     * @Route("/your-details/change-password", name="pa_profile_password_edit")
+     * @Template("AppBundle:Pa/Settings:profilePasswordEdit.html.twig")
+     **/
+    public function profilePasswordEditAction(Request $request)
```
elvisciotti: this is the same as the existing UserController::passwordEditAction(), except for the route and template.

I'd try to implement it by linking to the existing controller action, rendering the existing template, and just put some if/else (both controller and twig) based on what the user is (app.user.isDeputyPa() in twig, and $this->GetUser()->isDeputyPA() in the controller). What do you think ?

jackgoodby: No problemo

PR #728, opened 2017-08-01
Ddpb 1395 bugs

elvisciotti:


src/AppBundle/Entity/Report/Report.php
```diff
@@ -898,4 +898,19 @@ public function setWishToProvideDocumentation($wishToProvideDocumentation)
 
         return $this;
     }
+
+    /**
+     * @return string
+     */
+    public function getAttachmentNamePdf()
```
slizzio: As a refactor, this could be in the Abstract file type classes and called generateAttachmentName() which takes the extension for that file type ?

PR #732, opened 2017-08-04
Ddpb 1478 upload images support

slizzio:


src/AppBundle/Service/File/Checker/ClamAVChecker.php
```diff
@@ -56,45 +57,37 @@ public function checkFile(UploadableFileInterface $file)
...
-                    throw new RiskyFileException('Invalid PDF');
-            }
-
-            throw new RiskyFileException($fileScannerMessage);
+        switch($fileScannerCode) {
+            case 'AV_FAIL':
+                throw new VirusFoundException('Found virus in file');
+            case 'PDF_INVALID_FILE':
+            case 'PDF_BAD_KEYWORD':
+                throw new RiskyFileException('Invalid PDF');
```
elvisciotti: do we have a requirement for the message to display ? risky or invalid ? Symfony check for the MIME, so I guess invalid will be rare though

PR #742, opened 2017-08-14
Ddpb 1450

elvisciotti:


src/AppBundle/Command/DocumentCleanupCommand.php
```diff
@@ -3,32 +3,205 @@
...
         }
 
+        return $ret;
+    }
+
+    /**
+     * release lock.
+     *
+     * @param type $output
+     */
```
slizzio: @return param here would be useful, saves having to lookup redis docs

PR #752, opened 2017-08-30
Ddpb 1516 pa user address

jackgoodby:


slizzio:
Not merging as release is out this morning

PR #758, opened 2017-09-04
Ddpb 1404 update elements

Biggsen:


src/AppBundle/Resources/views/Components/Form/_known-date.html.twig
```diff
@@ -8,24 +8,28 @@
             {{ form_errors(element) }}
         </legend>
         {# day: show element or disable field depending on "showDay" setting #}
+        {% set errorClass = '' %}
+        {% if not element.vars.valid %}
+            {% set errorClass = ' form-control-error ' %}
+        {% endif %}
```
elvisciotti: you can also use the ternary here to shorten

{% set errorClass = element.vars.valid ? '' : ' form-control-error ' %}

PR #768, opened 2017-09-14
Ddpb 1551 homepage alignment

Biggsen:


tests/behat/features/pa/03-report/03-104/01-prepare-report-status-104.feature
```diff
@@ -3,7 +3,7 @@ Feature: Report 104 start
   @deputy @104
   Scenario: load app status to common sections completed, change type to 104 and check not submittable
     Given I load the application status from "102-common-sections-complete"
-    And I change the report 8 type to "104"
+    And I change the report 7 type to "104"
```
elvisciotti: 1053 will change this step and rely on case numbers instead of ID

PR #769, opened 2017-09-15
Ddpb 1609 content 104

Biggsen:


src/AppBundle/Resources/views/Report/Action/_answers.html.twig
```diff
@@ -1,6 +1,7 @@
 {% set translationDomain = "report-actions" %}
 {% trans_default_domain translationDomain %}
 {% set transOptions = {'%client%': report.client.firstname | e } %}
+{% set append104 = ('104' in report.type) ? '-104' : '' %}
```
elvisciotti: that will work for Lay and PA, including future joint reports that include 104 (103-4 and 102-4)

PR #773, opened 2017-09-21
DDPB-1606 fix declaration info missing in PDF

elvisciotti:


elvisciotti:
The problem was the PDF being generated from the data coming from the API, but the API wasn't yet updated to hold the declaration info. Solved by passing the updated report to the PDF generator. Storing declaration, submitting and sending the PDF is still a single "transaction"

PR #782, opened 2017-10-02
Ddpb 1615 fixes

slizzio:


src/AppBundle/Entity/Report/Report.php
```diff
@@ -1016,4 +1016,15 @@ public function getSectionsSettings()
...
+    /**
+     * Generates the translation suffic to use depending on report type,
+     *
+     * 10x followed by "-104" for HW, "-4" for hybrid report and nothing for PF report
+     * 
+     * @return string
+     */
+    public function get104TransSuffix()
+    {
+        return (strpos($this->getType(), '-4') > 0) ? '-4' : ($this->getType() === '104' ? '-104' : '');
```
elvisciotti: I'm 99.9% sure that the type comes as string, so that should be correct

elvisciotti: for me to summarise: 
- if it contains '-4' => joint report (both Lay and PA)
- if it's a pure 104 for Lay => than we return '104'

For 104 Pa (104-6), this method will return empty, is that what's needed ?

PR #783, opened 2017-10-02
Ddpb 1323 upload docs submitted report

slizzio:


src/AppBundle/Controller/Report/DocumentController.php
```diff
@@ -102,14 +103,20 @@ public function step1Action(Request $request, $reportId)
      */
     public function step2Action(Request $request, $reportId)
     {
-        $report = $this->getReportIfNotSubmitted($reportId, self::$jmsGroups);
-
-        if ( EntityDir\Report\Status::STATE_NOT_STARTED == $report->getStatus()->getDocumentsState()['state']
-           && 'yes' !== $report->getWishToProvideDocumentation()
```
elvisciotti: commenting here as not related to those changes: The ReportStatusService is normally used to tell the status, abstracting from what are the needed methods in the section (getWishToProvideDocumentation) in this case. No need to change, but to keep in mind and discuss in case we make future changes on those things

elvisciotti: In this case, if it's not started, the `getWishToProvideDocumentation`is supposed to be null, and there should be no need to check it. does it make sense ?

src/AppBundle/Controller/Report/DocumentController.php
```diff
@@ -207,6 +214,10 @@ public function deleteConfirmAction(Request $request, $documentId, $confirmed =
         /** @var EntityDir\Document $document */
         $document = $this->getDocument($documentId);
 
+        if ($document->getReportSubmission() instanceof EntityDir\Report\ReportSubmission) {
+            throw new \RuntimeException('Docuemnt already submitted and cannot be removed.');
```
elvisciotti: typo

src/AppBundle/Controller/Report/DocumentController.php
```diff
@@ -248,19 +260,70 @@ public function deleteConfirmedAction(Request $request, $documentId)
...
+     */
+    public function submitMoreConfirmedAction(Request $request, $reportId)
+    {
+        $report = $this->getReport($reportId, self::$jmsGroups);
+
+        // submit the report to generate the submission entry only
+        $response = $this->getRestClient()->put('report/' . $report->getId() . '/submit-documents', $report, ['submit']);
+
+        if ($response['reportId'] !== $report->getId()) {
+            throw new \RuntimeException('Unable to submit new documents');
```
elvisciotti: in what cases is this happening? is more likely to fail with an ecxeption from the endpoint ? if so, the previous `put(...)` will already throw an exception

src/AppBundle/Resources/views/Report/Document/_list_submitted.html.twig
```diff
@@ -0,0 +1,72 @@
...
+        {% if page == 'submitMoreDocumentsConfirm' and report.wishToProvideDocumentation == 'yes' %}
+            <a href="{{ path('report_documents', { 'reportId': report.id, 'from': 'summary'}) }}" class="bold-small move-quarter--up action-link behat-link-add">
+                <span class="icon icon-plus"></span>
+                {{ 'summaryPage.addButton' | trans }}
+            </a>
+        {% endif %}
+        </div>
+    </div>
+
+    <table class="check-your-answers push--bottom behat-region-new-document-list">
```
elvisciotti: maybe a partial template to reuse here ?  @Biggsen

tests/behat/features/deputy/03-report/01-102/21-add-more-documents.feature
```diff
@@ -0,0 +1,33 @@
...
+    Then the form should be valid
+    And each text should be present in the corresponding region:
+      | small.jpg        | new-document-list |
+
+    Then I click on "continue-to-submit"
+    Then the URL should match "report/\d+/documents/submit-more"
+    And each text should be present in the corresponding region:
+      | small.jpg        | new-document-list |
+    Then I click on "confirm-submit"
+    And I go to the URL previously saved as "your-reports"
```
elvisciotti: not needed if it's the last step

src/AppBundle/Controller/Report/DocumentController.php
slizzio: Yes thats a nonsensical statement. Might as well be if (true && false)

src/AppBundle/Controller/Report/DocumentController.php
slizzio: Yes this is true. Have removed. The current Id will not change since  its A PUT and already exists

tests/behat/features/deputy/03-report/01-102/21-add-more-documents.feature
slizzio: This is just to test its redirecting back to the correct place. Why would this not be needed? Have I got the step wrong?

src/AppBundle/Controller/Report/DocumentController.php
slizzio: Removed impossible if statement above

PR #792, opened 2017-10-09
Ddpb 1678 codeps sort order

jackgoodby:


src/AppBundle/Entity/Client.php
```diff
@@ -697,10 +695,25 @@ public function setClientContacts($clientContacts)
...
+        if (is_array($this->users) && count($this->users) > 0) {
+            foreach ($this->users as $user) {
+                if (!$user->getFirstname()) {
+                    $matches = [];
+                    preg_match('(^\w+)', $user->getEmail(), $matches);
+                    if (!empty($matches[0])) {
+                        $coDeps[strToLower($matches[0])] = $user;
+                    }
+                } else {
+                    $coDeps[strToLower($user->getFirstname())] = $user;
```
slizzio: There is a danger of this being overwritten and codeps going missing. Consider 2 co deps with the same first name. Maybe add a unique identifier?

jackgoodby: yeah of course - duh

PR #798, opened 2017-10-13
Ddpb 1628 admin document improvements

slizzio:


src/AppBundle/Resources/views/Admin/ReportSubmission/index.html.twig
```diff
@@ -107,13 +108,27 @@
                         </td>
                         <td class="width-fifth">
                             {% if reportSubmission.createdBy %}
+                                {% if reportSubmission.createdBy.isDeputyPa %}
+                                    {% set PAName = reportSubmission.createdBy.determinePublicAuthorityName %}
+                                        {% if PAName != '' and PAName != reportSubmission.createdBy.lastname %}
+                                            {{ reportSubmission.createdBy.fullname }}<br>
```
elvisciotti: this case is the same as line 121 and 122, so maybe this logic can be simplified?

slizzio: Yes it could be written more compactly. I didn't want to make a call to determinePublicAuthorityName unnecessarily (due to the regex). So whilst it would be more compact. it would be possibly harder to read

PR #806, opened 2017-10-17
DDPB-1704 review report heading fix

Biggsen:
looking for '104' in the string, instead of an exact '104' to match 104-6.

### DIFF
--- a/src/AppBundle/Resources/views/Report/Formatted/formatted_body.html.twig
+++ b/src/AppBundle/Resources/views/Report/Formatted/formatted_body.html.twig
@@ -8,7 +8,7 @@
 {% set transfers = report.getMoneyTransfers %}
 {% set mentalCapacity = report.getMentalCapacity %}
 {% set transOptions = {'%client%': report.client.firstname | e } %}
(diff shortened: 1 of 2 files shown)


src/AppBundle/Resources/views/Report/Formatted/formatted_body.html.twig
```diff
@@ -17,6 +17,9 @@
     {% include 'AppBundle:Css:formatted-report.css.twig' %}
 </style>
 
+<p>{{ report.get104TransSuffix }}</p>
```
slizzio: Is this meant to be here?

src/AppBundle/Resources/views/Report/Formatted/formatted_body.html.twig
```diff
@@ -17,6 +17,9 @@
...
+<p>{{ report.type }}</p>
```
slizzio: Are these meant to be here? Debug only?

Biggsen: Debug only. Gone.

Biggsen:
Not showing new commits.

PR #809, opened 2017-10-20
DDPB-1628-revert-template Reverted template and fixed pre match

slizzio:


src/AppBundle/Entity/User.php
```diff
@@ -923,8 +923,8 @@ public function isDeputy()
      */
     public function determinePublicAuthorityName()
```
Biggsen: We should remove the whole function. It won't be needed.

Biggsen:
Approved.

PR #811, opened 2017-10-23
DDPB-1329 mask STMP host in manage/availability in case of failures

elvisciotti:


src/AppBundle/Service/Availability/SmtpAvailability.php
```diff
@@ -11,17 +11,17 @@ class SmtpAvailability extends ServiceAvailabilityAbstract
...
+
+        $this->isHealthy = false;
+        $this->errors = '';
 
         try {
-            $transport = $container->get($this->transportKey); /* @var $transport \Swift_SmtpTransport */
             $transport->start();
             $transport->stop();
-
-            $this->isHealthy = true;
```
jackgoodby: When is $this->isHealthy every going to be true, if this is removed?

PR #814, opened 2017-10-23
DDPB-1168 registered email error message updated

Biggsen:


elvisciotti:
@Biggsen @jackgoodby the same message is displayed 3 times (+ once with a different phone number). Wasn't possible or convenient to re-use the same translation key and remove some content duplication ?

PR #821, opened 2017-10-27
Ddpb 1696 docs sent msg

Biggsen:


Biggsen:
Look like I accidentally branched off the trans-keys-fix branch (pageTitle work) instead of master - https://github.com/ministryofjustice/opg-digi-deps-client/pull/818

Let's have that merged in first before this.

PR #824, opened 2017-11-02
Ddpb 1707 monetary validation content

slizzio:


jackgoodby:
Not merged at present because code freeze

PR #840, opened 2017-11-22
Ddpb 1772

elvisciotti:


src/AppBundle/Controller/Admin/AdController.php
```diff
@@ -37,7 +37,6 @@ public function indexAction(Request $request)
                                  , new EntityDir\User()
                                  , [ 'options' => [ 'roleChoices'        => [EntityDir\User::ROLE_LAY_DEPUTY=>'Lay deputy'],
```
jackgoodby: or just make the third param a simple array: 

 ['roleChoices'        => [EntityDir\User::ROLE_LAY_DEPUTY=>'Lay deputy'],
'roleNameSetTo'      => EntityDir\User::ROLE_LAY_DEPUTY,
'roleNameEmptyValue' => null,]

src/AppBundle/Form/Ad/AddUserType.php
```diff
@@ -8,54 +8,30 @@
...
-
         $builder
             ->add('firstname', 'text')
             ->add('lastname', 'text')
-            ->add('roleName', 'choice', $roleNameOptions)
-            ->add('odrEnabled', 'checkbox', $odrEnabledOptions)
+            ->add('roleName', 'choice', [
+                    'empty_value' => null,
+                    'choices'     => $options['options']['roleChoices'],
+                    'data'        => $options['options']['roleNameSetTo'],
```
jackgoodby: If you're passing a single array of options these could be $options['roleChoices'], $options['roleNameSetTo'] etc

elvisciotti: yes, exactly what I'm planning to simplify after I write a behat test (that will guarantee I don't mess it up)

src/AppBundle/Controller/Admin/AdController.php
elvisciotti: I'll clean it up further after I write the test, I'll submit for a 2nd review later

PR #844, opened 2017-11-23
Ddpb 1723 incorrect report visibility

jackgoodby:


src/AppBundle/Resources/views/Pa/ClientProfile/overview.html.twig
```diff
@@ -316,7 +316,7 @@
     } %}
 
     {% include 'AppBundle:Pa/ClientProfile:_reports.html.twig' with {
-    reports: report.client.reports,
+    reports: report.client.getSubmittedReports(),
```
elvisciotti: report.client.submittedReports would be more in line

jackgoodby: sure

src/AppBundle/Entity/Client.php
```diff
@@ -716,4 +716,18 @@ public function getCoDeputies()
         }
         return array_values($coDeps);
     }
-}
+
+    /**
+     * @return array $submittedReports an array of submitted reports
+     */
+    public function getSubmittedReports()
```
elvisciotti: we get a list of submitted reports from lay_home page as well.
ReportController:84-91

What about some refactor to use this new method there?

a method getActiveReport  can be added alongside. And the the arsort can be left out in the controller

tests/behat/features/pa/03-report/01-102/01-report-edit-period.feature
```diff
@@ -1,5 +1,11 @@
 Feature: PA report
 
+  Scenario: PA does not see unsubmitted reports in the submitted reports section
+    Given I load the application status from "team-users-complete"
+    And I am logged in as "behat-pa1@publicguardian.gsi.gov.uk" with password "Abcd1234"
+    When I click on "pa-report-open" in the "client-1000014" region
+    Then I should see "No submitted reports" in the "client-profile-reports" region
```
elvisciotti: Is there already a test that assert that at least one submitted report is listed ?

src/AppBundle/Entity/Client.php
jackgoodby: Good call - Done!

src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -78,16 +78,9 @@ public function indexAction(Request $request)
...
-
-        $reportActive = null;
-        $reportsSubmitted = [];
-        foreach ($reports as $currentReport) {
-            if ($currentReport->getSubmitted()) {
-                $reportsSubmitted[] = $currentReport;
-            } else {
-                $reportActive = $currentReport;
-            }
+        $reportsSubmitted = $client ? $client->getSubmittedReports() : [];
```
elvisciotti: @jackgoodby the method doesn't do the arsort, so you are altering the logic here. You might need to re-add it on $reportsSubmitted (not inside the method, as it's not always wanted I think). Might be not neeed, but you need to check the usage to be sure

jackgoodby: arsort now line 82

PR #845, opened 2017-11-23
Ddpb 1733 balance page

Biggsen:


src/AppBundle/Resources/views/Report/Balance/balance.html.twig
```diff
@@ -5,24 +5,367 @@
...
+{% endif %}
+
+{% set transOptions = {
+    '%client%': report.client.firstname | e,
+    '%moreless%': moreless,
+    '%pathToAccounts%': path('bank_accounts', {'reportId': report.id})
+} %}
+
+{% if
+    reportStatus.giftsState['state'] != 'not-started' and
```
elvisciotti: this should stay in the backend

PR #849, opened 2017-12-01
Ddpb 1651 edit link content change

Biggsen:


src/AppBundle/Resources/views/Pa/ClientProfile/deleteConfirm.html.twig
```diff
@@ -27,7 +27,7 @@
             </tr>
             <tr>
                 <th scope="row">{{ (page ~ '.dateAdded') | trans }}</th>
-                <td>{{ note.getCreatedOn() | date('d F Y') }}</td>
+                <td>{{ note.getCreatedOn() | date('j F Y') }}</td>
```
elvisciotti: equivalent to note.createdOn

twig automatically add  "get" and use it as a method. Just to align our knowledge, no need to change

PR #857, opened 2017-12-11
Ddpb 1750 static info presentation

Biggsen:


src/AppBundle/Resources/views/User/_form_fields.html.twig
```diff
@@ -13,17 +15,53 @@
...
         {{ form_widget(form.lastname, {'type': 'hidden'}) }}
-        <p class="push--bottom">
+        <p class="flush--bottom">
             <strong class="bold">{{ 'form.editYourDetails.controls.lastname.label' | trans({}, translationDomain) }}: </strong>
             {{ user.lastname }}
         </p>
     {% endif %}
 {% endif %}
 
+{% if form.address1 is defined %}
```
elvisciotti: it could be a single if with "and"

PR #859, opened 2017-12-14
DDPB-1840 Mobile links broken hotfix

jackgoodby:
Reverts some of the frontend changes made on 30th Nov 2017 in association with DDPB-1758 to reinstate an <a> tag around the whole subsection. Otherwise the page is broken with the “Edit” links set to display:none (mobile view)

### DIFF
--- a/src/AppBundle/Resources/views/Report/Report/_subsection.html.twig
+++ b/src/AppBundle/Resources/views/Report/Report/_subsection.html.twig
@@ -14,6 +14,9 @@
 {% endif %}
 
 <div id="{{ subSection }}-sub-section" class="section-list-item">
+    {% if allowEdit == true %}
+        <a class="behat-link-edit-{{ subSection }}" id="edit-{{ subSection }}" href="{{ linkToSubSection }}" aria-label="{{ (subSection ~ '.edit') | trans }}">
+    {% endif %}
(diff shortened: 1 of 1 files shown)


src/AppBundle/Resources/views/Report/Report/_subsection.html.twig
```diff
@@ -23,33 +26,28 @@
...
-                {{ 'view' | trans({}, 'common') | capitalize }}
-            {% else %}
-                {% if state.state == 'not-started' %}
-                    {{ 'start' | trans({}, 'common') | capitalize }}
-                {% else %}
-                    {{ 'edit' | trans({}, 'common') | capitalize }}
-                {% endif %}
-            {% endif %}
-        </a>
+    <span  class="edit-link {% if allowEdit == false %}disabled{% endif %}" {% if allowEdit == true %}href="{{ linkToSubSection }}"{% endif %} >
```
slizzio: here ^^

jackgoodby: This is consistent with the revert - every version of this file had a href in the <span> tag. I don't agree with it either, but I was just reverting

slizzio: Discussed and this line is reverting the file back to how it was

PR #860, opened 2017-12-15
Ddpb 1559

elvisciotti:


jackgoodby:
@elvisciotti Not sure I agree with commenting out all those 'And I save the page'... why not just remove them?

elvisciotti:
@jackgoodby  agreed. I commented them out first to see if resulted into faster execution on jenkins. It does, and I can remove them now. Those checkpoints were useful to have screenshots of the site for testers, but we don't use them, so I'll remove them all

tests/behat/bootstrap/DbTrait.php
```diff
@@ -13,8 +13,9 @@
     public static function iSaveTheApplicationStatusInto($status)
     {
         $sqlFile = self::getSnapshotPath($status);
-        @unlink($sqlFile);
-        exec('pg_dump ' . self::$dbName . " --clean --inserts | sed '/EXTENSION/d' > {$sqlFile}", $output, $return);
+        // manual data
+        exec('echo "SET client_min_messages TO WARNING; truncate dd_user, dd_team, casrec, client cascade;" > '.$sqlFile);
+        exec('pg_dump ' . self::$dbName . "  --data-only  --inserts --exclude-table='migrations' | sed '/EXTENSION/d' >> {$sqlFile}", $output, $return);
         if (!file_exists($sqlFile) || filesize($sqlFile) < 100) {
```
slizzio: could possibly remove this check for file exists to speed things up further as the filesize will still fail if doesnt exist

slizzio: in fact lose the entire if ?

elvisciotti: I'd not remove it, I'd trigger a warning if not existing although the eception would be thrown anyway. Also, takes a few ms to run that command, nothing compare to the 15 minutes to run the whole suite, and therefore I'd not even bother making the change and make sure I didn't alter the logic. Maybe we can do that on master next time  we change it

slizzio: Yes, fair enough, I dont think it would even get that far as the exec statement above would fail to write to the file.

PR #862, opened 2017-12-18
DDPB-1776 PDF bad/good balance content update

Biggsen:


jackgoodby:
Approved - self-merge when master becomes available

PR #873, opened 2018-01-05
Ddpb 1759 - PDF summary info in PDF for OPG

elvisciotti:


src/AppBundle/Resources/views/Report/Formatted/_report_summary.html.twig
```diff
@@ -0,0 +1,105 @@
...
+{% set transOptions = {
+    '%moreless%': report.totalsOffset < 0 ? "less" : "more",
+} %}
+
+<div class="section" id="report-summary">
+    <h2 class="section-heading">Report summary</h2>
+
+    {% if report.hasSection('actions') %}
+        {{ macros.answerYesNo({
+            question: 'Significant Decisions',
```
jackgoodby: Could / should we be using a translation file for content?

src/AppBundle/Entity/Report/Traits/ReportAssetTrait.php
```diff
@@ -53,6 +66,27 @@ public function getAssetsTotalValue()
         return $this->assetsTotalValue;
     }
 
+    /**
+     * @param string $type asset|cash|other
```
jackgoodby: property|cash|other

src/AppBundle/Entity/Report/Traits/ReportAssetTrait.php
```diff
@@ -9,6 +9,19 @@
 
 trait ReportAssetTrait
 {
+    /**
+     * Titles matching this will be included in the count for "Cash" in summary page
+     * Note: it relies on the translation (see report-assets.en.yml form.choices) for historical reasons
+     *
+     */
+    private static $cashAssetTitles = [
```
jackgoodby: Wouldn't work with non-english ;)

elvisciotti: the db only stores titles, due to the fact that types were added later. Fixing that would need a big data migration that was never prioritised. I'd say we do that next time we work on assets, that never changed for over 1 year

elvisciotti: it won't work if we rename the translation. I'll add a note into the translation file itself as a comment, so that we won't forget to update the code as well

src/AppBundle/Resources/views/Report/Formatted/_report_summary.html.twig
elvisciotti: everything under src/AppBundle/Resources/views/Report/Formatted is very little translated. Not sure we want to plan to refactor that as in the future we'll send the data directly to Sirius and the PDF will die at a certain point

PR #877, opened 2018-01-11
Ddpb 1792 tweaks

Biggsen:


src/AppBundle/Resources/views/Report/MoneyOut/step.html.twig
```diff
@@ -28,6 +28,8 @@
     {% if step == 2 %}
         {% if transaction.group == "fees" %}
             {{ groupString }}
+        {% elseif transaction.group == "major-purchases" %}
+            major purchase
```
elvisciotti: @Biggsen shouldn't that use {{ groupString }} (with a "lower" filter if need) ?

PR #892, opened 2018-01-29
Ddpb 1881 multiple download ui

Biggsen:


tests/behat/bootstrap/FileTrait.php
```diff
@@ -35,6 +35,19 @@ public function thePagecontentShouldBeZipContainingFilesChecksum(TableNode $tabl
...
 
+                case 'exactFileName+filesize':
+                    $expectedSize = $value;
+
+                    exec("unzip -l $tmpFile | grep -E \"{$file}\" ", $lines);
+                    if (empty($lines)) {
+                        throw new \RuntimeException("File matching $file not found in ZIP file");
+                    }
+                    $sizeBytes = array_shift(array_filter(explode(' ', $lines[0])));
+                    if ($sizeBytes <> $value) {
```
elvisciotti: minor: I guess ZIP files will always have the same size with the same files, but in a next PHP version they might change the ZIP file default compression and break this code. For PDF I've used a min size, see "@Then the :index email should contain a PDF of at least :minsizekb kb". Maybe something similar ?

slizzio: Ah I see, yes that makes sense. I was wondering what minSize was for

PR #903, opened 2018-02-12
Ddpb 1818 add user

slizzio:


src/AppBundle/Resources/views/Pa/Team/list.html.twig
```diff
@@ -16,7 +16,8 @@
 {% block pageContent %}
     {# Introduction text #}
     <p class="text">
-        {% if (app.user.roleName == 'ROLE_PA_NAMED' or app.user.roleName == 'ROLE_PA_ADMIN') %}
+        {% if ( app.user.roleName == 'ROLE_PA_NAMED' or app.user.roleName == 'ROLE_PA_ADMIN' or
```
elvisciotti: maybe we can use an User method here, is isGranted ?

tests/behat/features/deputy/06-static-pages/03-terms.feature
```diff
@@ -1,4 +1,4 @@
-Feature: Terßms and Conditions
```
elvisciotti: German version :)

PR #908, opened 2018-02-14
Ddpb 1888 ndr jms fix

elvisciotti:


src/AppBundle/Controller/AbstractController.php
```diff
@@ -81,6 +81,7 @@ public function getReportsIndexedById(Client $client, $groups = [])
     public function getReport($reportId, array $groups = [])
     {
         $groups[] = 'report';
+        $groups[] = 'report-client';
         $groups[] = 'client';
```
slizzio: should this be removed?

elvisciotti: report-client is only triggering the relationships Report::getClient(), but it'd retrieve back an empty Client. The "client" group will add the client fields as well, so report.client.firstName will be available that way

In other words, relationships need to be unique so that we can control them from the client. A combination of groups report+report-client+client-id for example would retrieve back report.client.id but not report.client.firstname.

slizzio: ah yes cool

PR #913, opened 2018-02-19
Ddpb 1828 view current report

slizzio:


src/AppBundle/Resources/views/Report/MoneyOut/start.html.twig
```diff
@@ -13,9 +13,9 @@
     '%paDeputyExpensesPath%': path('pa_fee_expense', {reportId: report.id}),
     '%giftsPath%': path('gifts', {reportId: report.id})
 } %}
-{% if report.hasSection('deputyExpenses') %}
```
elvisciotti: deputyExpenses is for lay deputies, and should stay. Why is it removed /

slizzio: This if statment was used to set a variable appendPA to be nothing ('') if the report contained deputyExpenses section. I have replaced that variable by defaulting it to nothing. Hence no if statement needed. It was breaking the prof tests because appendPA variable was not set to anything:
Actual diff has replaced this:
{% if report.hasSection('deputyExpenses') %}
    {% set appendPA = '' %}
{% elesif report.hasSection('paDeputyExpenses') %}
    {% set appendPA = 'PA' %}
{% endif %}
with this:
{% set appendPA = '' %}
{% if report.hasSection('paDeputyExpenses') %}
    {% set appendPA = 'PA' %}
{% endif %}
To ensure appendPA is set for *all* report types

elvisciotti: ok ta, understood when I saw the fullcode

PR #916, opened 2018-03-06
Incomplete reports

elvisciotti:


src/AppBundle/Resources/translations/common.en.yml
```diff
@@ -26,6 +27,7 @@ manage: Manage
 annualDeputyReport: Annual deputy report
 currentDeputyReport: Current deputy report
 submittedReports: Submitted reports
+unsubmittedReport: Incomplete deputy report
```
slizzio: hmmm key is a bit confusing here

src/AppBundle/Service/Client/RestClient.php
```diff
@@ -280,6 +280,10 @@ public function apiCall($method, $endpoint, $data, $expectedResponseType, $optio
             $options['body'] = $this->toJson($data, $options);
         }
 
+//        if ($method=='put') {
+//            echo "<pre>";\Doctrine\Common\Util\Debug::dump($options, 4);die;
+//        }
+
```
slizzio: To remove or remain?

src/AppBundle/Resources/translations/common.en.yml
elvisciotti: in the backend, incomplete means a report not submitted, but this new functionality calls them "incomplete" in the UI. I and Biggs decided to keep the naming consistent

src/AppBundle/Service/Client/RestClient.php
elvisciotti: ups. well spotted. I'll remove this

PR #917, opened 2018-03-12
Ddpb 1984 prof current fees

elvisciotti:


src/AppBundle/Entity/Report/ProfServiceFee.php
```diff
@@ -0,0 +1,291 @@
...
+        'appointment' => false,
+        'conveyancing' => false,
+        'tax-returns' => false,
+        'trust-applications' => false,
+//        'litigation' => true,
+//        'specialist-advice' => true,
+//        'statutory-wills' => false,
+        'other-costs' => false
+    ];
+
```
Biggsen: @elvisciotti Let's just delete the ones we don't need. Also why is the same list of service types in the API?

elvisciotti: I cleaned things up

PR #942, opened 2018-04-04
Ddpb 2010 update s3 purge tag

slizzio:


src/AppBundle/Service/File/Storage/S3Storage.php
```diff
@@ -98,8 +109,65 @@ public function store($key, $body)
...
+     *
+     * @param $level
+     * @param $message
+     */
+    private function log($level, $message)
+    {
+        //echo $message."\n"; //enable for debugging reasons. Tail the log with log-level=info otherwise
+
+        $this->logger->log($level, $message, ['extra' => [
+            'service' => 'documents-service',
```
slizzio: Change service to s3-storage

PR #944, opened 2018-04-09
Ddpb 2045 account dropdown

slizzio:


src/AppBundle/Controller/Report/BankAccountController.php
```diff
@@ -211,9 +211,24 @@ public function deleteAction(Request $request, $reportId, $accountId)
                 'notice',
                 'Bank account deleted'
             );
+        } catch (RestClientException $e) {
```
elvisciotti: just for curiosity, what happens if the API fails for another RestClientException  (not a BusinessRulesException one ?

src/AppBundle/Form/Report/DeputyExpenseType.php
```diff
@@ -20,8 +23,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'grouping' => true,
                 //'error_bubbling' => true,  // keep (and show) the error (Default behaviour). if true, error is los
                 'invalid_message' => 'expenses.singleExpense.notNumeric',
-            ])
-            ->add('save', 'submit');
+            ]);
+
+        if ($options['user']->getRoleName() == User::ROLE_LAY_DEPUTY) {
```
elvisciotti: Are the forms only or Lay deputies?

src/AppBundle/Controller/Report/BankAccountController.php
slizzio: Tested, nothing happens on frontend. Log entry made. Ive rethrown the exception if data is not set. Handled in same way as any other RestClientException.
I cant catch the BusinessRuleException as all exceptions are converted to RestClientException. Does this need to change?

PR #951, opened 2018-04-12
Ddpb 2049 zip

elvisciotti:


elvisciotti:
bugs:
- file name collision when the same report gets re-submitted with more documents
- response was not returned, so the HTML page content was appended to the ZIP (bothering windows ZIP file in combination with the previous bug)

PR #958, opened 2018-04-19
Ddpb 2054 report submission service

slizzio:


src/AppBundle/Service/ReportSubmissionService.php
```diff
@@ -0,0 +1,108 @@
...
+    private $wkhtmltopdf;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * DocumentService constructor.
+     * @param S3Storage       $s3Storage
```
elvisciotti: to update

PR #963, opened 2018-04-24
Ddpb 2054 accounts csv

slizzio:


src/AppBundle/Service/CsvGeneratorService.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Generates a description. Expenses and Gifts have an 'explanation' property,
+     * Money transactions have a description property.
+     *
+     * @param $transaction
+     * @return string
+     */
+    private function generateDescription($transaction)
+    {
+        if (method_exists($transaction, 'getDescription')) {
```
elvisciotti: alternative could be using interfaces hasDescription, hasExplnation, add use "instance of" on those. Less risk of typos and string mismatch

PR #990, opened 2018-06-05
Ddpb 1863 translations missing

elvisciotti:


src/AppBundle/Resources/views/Admin/Index/index.html.twig
```diff
@@ -36,7 +36,7 @@
     </div>
 
     <div class="form-group">
-        {{ form_submit(form.search, 'Search', {
+        {{ form_submit(form.search, '', {
             'labelText': 'Search'}) }}
```
slizzio: I think this will override everything including the translation key?

slizzio: Confirmed ok to merge

PR #1000, opened 2018-06-18
update to symfony3.4

elvisciotti:


app/config/services_api.yml
```diff
@@ -2,16 +2,16 @@ services:
...
-            base_url: "%api_base_url%"
-            defaults:
-                # force HTTP 1.0 in order to disable HTTP1.1-chunking corrupting JSON decoding
-                version: "1.0"
-                headers:
-                    Content-Type: application/json
-                verify: false
-                timeout: 20
-                allow_redirects: false
+          $config:
```
slizzio: is the $ intended?

src/AppBundle/Controller/Org/ClientContactController.php
```diff
@@ -68,7 +68,7 @@ public function editAction(Request $request, $id)
 
         $this->denyAccessUnlessGranted('edit-client-contact', $client, 'Access denied');
 
-        $form = $this->createForm(new FormDir\Org\ClientContactType($this->get('translator')), $clientContact);
+        $form = $this->createForm(FormDir\Org\ClientContactType::class, $clientContact);
```
slizzio: q: Does this still form translate ok?

src/AppBundle/Form/Report/ProfServiceFeeType.php
```diff
@@ -17,21 +17,11 @@ class ProfServiceFeeType extends AbstractType
...
-    {
-        $this->serviceTypeIds = $serviceTypeIds;
-    }
-
     private function getServiceFeeTypes()
     {
         $ret = [];
 
-        foreach ($this->serviceTypeIds as $serviceTypeId => $hasMoreInfo) {
+        foreach (ProfServiceFee::$serviceTypeIds as $serviceTypeId => $hasMoreInfo) {
```
slizzio: where do the serviceTypeIds come from, looks like they were removed above?

src/AppBundle/Controller/Org/ClientContactController.php
elvisciotti: yes. the translator wasn't used inside (previous refactor I think). It's generally not needed inside forms, as symfony translates choice values too (in symfony 3, they are now the keys, no the valeus of the "choices" array)

src/AppBundle/Form/Report/MoneyTransferType.php
```diff
@@ -19,22 +19,22 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $banks = [];
         foreach ($options['banks'] as $bank) {
             /* $var $bank \AppBundle\Entity\Report\BankAccount */
-            $banks[$bank->getId()] = $bank->getNameOneLine();
+            $banks[$bank->getNameOneLine()] = $bank->getId();
```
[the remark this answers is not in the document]
elvisciotti: I thought the same, but that method returns the account number too (in addition to type and account name), so there should be no duplicates

tests/phpunit/Service/Client/RestClientTest.php
```diff
@@ -447,7 +447,7 @@ public function testGetHistory()
             ->shouldReceive('getParameter')->with('kernel.debug')->andReturn(true)
             ->getMock();
 
-        $this->endpointResponse = m::mock('GuzzleHttp\Message\Response');
+
```
slizzio: looks like an abundance of blank lines here.

PR #1011, opened 2018-06-28
Ddpb 2162 rename final decision

Biggsen:


src/AppBundle/Resources/views/Admin/Client/Report/Formatted/checklist_formatted_body.html.twig
```diff
@@ -280,6 +280,11 @@
         <div class="section">
             <h2 class="section-heading">{{ (page ~ '.lodgingSummary.heading') | trans({}, 'admin-checklist') }}</h2>
 
+            <div class="labelvalue push-half--top">
+                <h3 class="label question bold">{{ (page ~ '.form.lodgingSummary.label') | trans({}, 'admin-checklist') }}</h3>
```
elvisciotti: @Biggsen admin-checklist could be set a default translation domain

PR #1014, opened 2018-06-29
DDPB-2161 documents on checklist. assets trans fix.

Biggsen:
…n fix.

### DIFF
--- a/src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
+++ b/src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
@@ -62,6 +62,34 @@
         </p>
     {% endif %}
 
+    {% if report.deputydocuments %}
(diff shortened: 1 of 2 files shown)


src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
```diff
@@ -62,6 +62,34 @@
         </p>
     {% endif %}
 
+    {% if report.deputydocuments %}
```
slizzio: is this correct? I thought it was report.documents?

elvisciotti: I'd name this getters deputyDocuments, more in line with the naming we use. Twig is case insensitive, but  we should respect the case as much as possible to make it easier to find&replace when needed

PR #1017, opened 2018-07-06
Ddpb 2223 casrec report csv

slizzio:


src/AppBundle/Service/CsvGeneratorService.php
```diff
@@ -62,6 +62,33 @@ public function generateTransactionsCsv(ReportInterface $report)
...
+     */
+    public function generateReportSubmissionsCsv($records)
+    {
+        $this->logger->info('Generating Report submissions CSV : ');
+
+        // initialize temporary fp
+        $this->fd = fopen('php://temp/maxmemory:1048576', 'w');
+        if ($this->fd === false) {
+            $this->logger->error('Failed to open Temporary file');
+            die('Failed to open temporary file');
```
elvisciotti: exception ?

slizzio: yep, moved to new private method and throw new Exception

PR #1028, opened 2018-07-20
Ddpb 2189 pa checklist

slizzio:


src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
```diff
@@ -63,7 +63,8 @@
     {% endif %}
 
     {# Attached documents #}
-    {% if report.deputydocuments %}
+
+    {% if report.deputyDocuments is defined and (report.deputyDocuments is not empty)%}
```
elvisciotti: getter should always be defined I think from twig

src/AppBundle/Resources/views/Admin/Client/Report/partials/_declaration.html.twig
```diff
@@ -35,7 +35,11 @@
                         </tr>
                         <tr>
                             <td>{{ 'deputyName' | trans({}, 'report-declaration') }}</td>
-                            <td>{{ report.submittedBy.fullname }}</td>
+                            {% if report.submittedBy is defined and report.submittedBy is not empty %}
```
elvisciotti: same

PR #1043, opened 2018-08-10
Ddpb 2252 checklist form title

slizzio:


src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
```diff
@@ -38,6 +38,10 @@
         {{ 'courtOrderNumber' | trans({}, 'common') }}:
         <span class="bold behat-region-case-number">{{ report.client.caseNumber }}</span>
     </p>
+    <p>
+        {{ 'reportType' | trans({}, 'common') }}:
+        <span class="bold behat-region-case-number">{{ ('reportTitles.' ~ report.reportTitle) | trans({}, 'report') }}</span>
```
elvisciotti: behat-region to remove

slizzio: yes good spot, removing

PR #1049, opened 2018-08-29
DDPB-2300 added 'hidden' class. fixed the spacing as well.

Biggsen:


Biggsen:
No dates showing under the first deputy and client info question:
image

PR #1051, opened 2018-09-28
DDPB-2310-dashboard-performance

elvisciotti:


src/AppBundle/Entity/Report/Report.php
```diff
@@ -657,7 +657,8 @@ public function isValidDateRange(ExecutionContextInterface $context)
 
     /**
      * Return true when the report is Due (today's date => report end date).
-     *
+     * //TODO consider using the API value instead of recalculating
```
slizzio: I don't think this is necessary is it? In any given day a report is either due or not due based on the current day. Should be a simple comparison like what you've coded up. Unless i've missed something?

elvisciotti: The API does the calculation already, since it's not straightforward (need to transform one date into midnight), we could use a virtual property and take it from the API. For sure the code below is needed as it fixes the bug mentioned in the team channel earlier

PR #1055, opened 2018-10-29
DDPB-2242 Make case number not editable for Lay deputy Client details form

digitalali-moj:
Please note that I have removed something from client form object for non-org - I don't think it's shared with anything else but please double check for me


digitalali-moj:
Done a better way in another PR

PR #1058, opened 2018-10-30
DDPB-2242 quick test of making case number readonly in form object

digitalali-moj:


digitalali-moj:
Done a better way in other PR

PR #1062, opened 2018-11-02
DDPB-2294 HW checklist

digitalali-moj:


src/AppBundle/Resources/views/Admin/Client/Report/checklist.html.twig
```diff
@@ -110,18 +111,30 @@
...
     {% include 'AppBundle:Admin/Client/Report/partials:_contact-with-client.html.twig' %}
 
+    {# Health & Lifestyle #}
+    {% if report.hasSection('lifestyle') %}
+        {% include 'AppBundle:Admin/Client/Report/partials:_health-and-lifestyle.html.twig' %}
+    {% endif %}
+
     {# Assets and debts #}
-    {% include 'AppBundle:Admin/Client/Report/partials:_assets-and-debts.html.twig' %}
+    {% if report.hasSection('assets') or report.hasSection('debts') %}
```
slizzio: Might be a discrepancy in the Report Variations page, but the 102-4-5 has debts section but not assets. Something to check is correct?

elvisciotti: It was an error in the matrix. The API config code was correct. Fixed the wiki page

PR #1063, opened 2018-11-05
Ddpb 1710 document label admin

digitalali-moj:


digitalali-moj:
Done on another branch

PR #1064, opened 2018-11-06
Ddpb 1710 reloaded

digitalali-moj:


src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -127,6 +127,17 @@ public function setDocuments($documents)
         return $this;
     }
 
+    /**
+     * @return bool
+     */
+    public function hasReportPdf()
+    {
+        foreach ($this->documents as $document) {
+            if($document->isReportPdf()) return true;
```
elvisciotti: we use the syntax:

if ($document->isReportPdf()) { 
 return true; 
}

PR #1065, opened 2018-11-06
DDPB-1867 NDR submit page same as annual report

digitalali-moj:


src/AppBundle/Controller/Ndr/NdrController.php
```diff
@@ -248,9 +249,45 @@ public function submitConfirmationAction(Request $request, $ndrId)
...
+     * @Template("AppBundle:Report:Report/submitFeedback.html.twig")
+     */
+    public function submitFeedbackAction($ndrId)
+    {
+        $client = $this->getFirstClient(self::$ndrGroupsForValidation);
+        $ndr = $client->getNdr();
+        if ($ndr->getId() != $ndrId) {
+            throw new \RuntimeException('Not authorised to access this Report');
+        }
+        $ndr->setClient($client);
```
slizzio: why are we setting the client to the ndr? surely its already set and known if the ndr has been submitted?

PR #1073, opened 2018-11-16
DDPB-2238 remove IE code

digitalali-moj:


digitalali-moj:
Rebased against master

PR #1074, opened 2018-11-16
DDPB-2312 Added additional logging

slizzio:


src/AppBundle/Controller/UserController.php
```diff
@@ -180,19 +180,32 @@ public function detailsAction(Request $request)
...
+                $emailAddress = $user->getEmail();
+                $disguisedEmail = '***' . substr($emailAddress, 3);
+                $logger->debug('Reset password request for : ' . $emailAddress);
                 /* @var $user EntityDir\User */
                 $user = $this->getRestClient()->userRecreateToken($user->getEmail(), 'pass-reset');
+                if (empty($user)) {
+                    $logger->warning('Email ' . $emailAddress . ' not found');
+                }
+
+                $logger->debug('Sending reset email to ' . $disguisedEmail);
```
elvisciotti: kibana should only log warnings going up, so only the 
'Email ' . $emailAddress . ' not found'
will be logged
is that wanted?

slizzio: is it not possible to change Kibana to show debug messages?

slizzio: Not sure whats wanted... ticket just says add extra logging. For me I just wanted the email address being requested to identify a possible pattern

src/AppBundle/Controller/UserController.php
```diff
@@ -180,19 +180,32 @@ public function detailsAction(Request $request)
...
+
         $user = new EntityDir\User();
         $form = $this->createForm(FormDir\PasswordForgottenType::class, $user);
 
         $form->handleRequest($request);
         if ($form->isValid()) {
             try {
+                $emailAddress = $user->getEmail();
+                $disguisedEmail = '***' . substr($emailAddress, 3);
+                $logger->warning('Reset password request for : ' . $emailAddress);
```
elvisciotti: this shows the real email address

src/AppBundle/Controller/UserController.php
elvisciotti: symfony creates hundreds of debug messages for each request and it'd crash kibana. There is a finger crossed handler to log debug messages, but tricky to enable and test for all the environments. In this case easier to just keep the "warning" level. Just one with the disguised email saying if it was found or not. See the comment about the real mail being logged

PR #1078, opened 2018-11-22
DDPB-2327 Remove GSI from translations/views (f/e only)

digitalali-moj:


digitalali-moj:
Removed gsi from views and email template translations.

Also removed from CSVs and Behat tests. Put it back for admin account, since otherwise can't login.

After several behat test runs and adjustments it seems to be all green now. (TBC)

PR #1103, opened 2019-01-22
DDPB-2383: Fixes PDF report hardcoded Client and balance total

aiesoftware:
https://opgtransform.atlassian.net/browse/DDPB-2383

### DIFF
--- a/src/AppBundle/Resources/views/Report/Formatted/Accounts/_final_summary.html.twig
+++ b/src/AppBundle/Resources/views/Report/Formatted/Accounts/_final_summary.html.twig
@@ -16,7 +16,7 @@
             </tr>
             <tr>
                 <td>Money out:</td>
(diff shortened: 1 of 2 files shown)


src/AppBundle/Resources/views/Report/Formatted/Accounts/_final_summary.html.twig
```diff
@@ -16,7 +16,7 @@
             </tr>
             <tr>
                 <td>Money out:</td>
-                <td class="grey numeric">&pound;{{ report.moneyOutTotal | money_format}}</td>
+                <td class="grey numeric">&pound;{{ (report.moneyOutTotal + report.profDeputyTotalCosts) | money_format}}</td>
```
slizzio: this will need a check on hasSection otherwise when we feature flag to disable the section, this will break

PR #1147, opened 2019-03-06
DDPB 2463 Page timeout on user edit page

aiesoftware:


src/AppBundle/Controller/Admin/IndexController.php
```diff
@@ -179,7 +178,7 @@ public function editUserAction(Request $request)
 
                     $request->getSession()->getFlashBag()->add('notice', 'Your changes were saved');
 
-                    $this->redirect($this->generateUrl('admin_editUser', ['what' => 'user_id', 'filter' => $user->getId()]));
```
slizzio: thanks for getting rid of the 'what' filter. Never liked it.. :)

src/AppBundle/Resources/views/Admin/Index/editUser.html.twig
```diff
@@ -19,8 +19,7 @@
 
     {% set numberOfClients = user.clients | length %}
     {% set firstClient = numberOfClients > 0 ? (user.clients | first) : null %}
-    {% set reports = firstClient ? firstClient.reports : [] %}
-    {% set reportsCount = reports | length %}
+    {% set reportsCount = firstClient ? firstClient.totalReportCount : 0 %}
```
slizzio: never liked the firstClient being used either. But that can be looked at another time

PR #1151, opened 2019-03-08
Feature/DDPB-2458 - Add general management figure to summary, Preview page and PDF

acsauk-mojd:
Comparing to the previous branch in this piece of work (`Feature/DDPB-2458`) until its merged. Once `Feature/DDPB-2458` is merged I'll update the base branch to `feature-105`.

I don't see any tests related to the actual contents of the PDF preview so I've not added any assertions for that - if anyone thinks theres value in that let me know I can add something in.

### DIFF
--- a/src/AppBundle/Resources/views/Report/ProfDeputyCostsEstimate/_answers.html.twig
+++ b/src/AppBundle/Resources/views/Report/ProfDeputyCostsEstimate/_answers.html.twig
@@ -37,7 +37,7 @@
                     {{ 'breakdown.form.profDeputyCostsEstimateManagementCost.label' | trans }}
                 </td>
                 <td>
-                    £{{ (report.profDeputyManagementCostAmount) | money_format }}
+                    £{{ (report.getProfDeputyManagementCostAmount) | money_format }}
                 </td>
                 {% if not hideEditLink  %}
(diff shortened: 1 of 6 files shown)


src/AppBundle/Resources/views/Report/ProfDeputyCostsEstimate/_answers.html.twig
```diff
@@ -37,7 +37,7 @@
                     {{ 'breakdown.form.profDeputyCostsEstimateManagementCost.label' | trans }}
                 </td>
                 <td>
-                    £{{ (report.profDeputyManagementCostAmount) | money_format }}
+                    £{{ (report.getProfDeputyManagementCostAmount) | money_format }}
```
slizzio: Just for info, you don't need to specify 'get', trig automatically calls a get method based on the property you want

acsauk-mojd: Ah nice, wasn't aware of that. I think I added this during debugging why it wasn't pulling through until I realised the property name wasn't matching from the api. I'll switch back to `profDeputyManagementCostAmount ` for consistency 👍

PR #1158, opened 2019-03-13
Feature/DDBP-2456 - Add client name and Case ref to submitted email

acsauk-mojd:
I've had an initial go at drying some of the translation files out where it made sense as there was a lot of repeated markup - not sure if theres a better way to achieve this?


src/AppBundle/Service/Mailer/MailFactory.php
```diff
@@ -259,8 +259,9 @@ public function createReportSubmissionConfirmationEmail(EntityDir\User $user, En
         $viewParams = [
             'submittedReport' => $submittedReport,
             'newReport'       => $newReport,
-            'deputyFirstName' => $user->getFirstname() . ' ' . $user->getLastname(),
-            'newReport'       => $newReport,
```
slizzio: why was this removed?

src/AppBundle/Resources/views/Email/report-submission-confirm.html.twig
```diff
@@ -4,18 +4,26 @@
...
+    '%homepageUrl%': homepageUrl,
+    '%fullClientName%': fullClientName,
+    '%caseNumber%': caseNumber
 } %}
 
 {% block emailHTMLHeaderTitle %}{{ ('default.htmlHeaderTitle') | trans }}{% endblock %}
 
 {% block content %}
 
+    {% if recipientRole != 'default' %}
```
slizzio: Professionals and PAs only?

acsauk-mojd: Yeah - confirmed in the ticket. From what I could see in `MailFactory::getRecipientRole()`, anything other than PA or Professional will be set to `default` so hopefully this should be enough?

slizzio: Yes, checked and confirmed, default it seems to be.

slizzio: Confirmed

acsauk-mojd:
Closing in favour of https://github.com/ministryofjustice/opg-digi-deps-client/pull/1163 as jenkins cant build from branches that contain a `/`.

PR #1161, opened 2019-03-14
DDPB-2387 Remove duplicate flash notice on checklist

digitalali-moj:
Removed duplicate code and put condition to not show if notice
already there for checklist saved. Probably did something silly
as it's b/e code. But thought I'd give it a shot.

### DIFF
--- a/src/AppBundle/Controller/Admin/Client/ReportController.php
+++ b/src/AppBundle/Controller/Admin/Client/ReportController.php
@@ -162,15 +162,17 @@ class ReportController extends AbstractController
             $checklist->setButtonClicked($buttonClicked->getName());
         }
         if ($form->isValid($buttonClicked)) {
+            $noticeAlreadyThere = $request->getSession()->getFlashBag()->has('notice');
             if (!empty($checklist->getId())) {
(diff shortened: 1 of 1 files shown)


src/AppBundle/Controller/Admin/Client/ReportController.php
```diff
@@ -162,15 +162,17 @@ public function checklistAction(Request $request, $id)
...
                 $this->getRestClient()->put('report/' . $report->getId() . '/checked', $checklist, [
                     'report-checklist', 'checklist-information'
                 ]);
-                $request->getSession()->getFlashBag()->add('notice', 'Lodging checklist saved');
             } else {
                 $this->getRestClient()->post('report/' . $report->getId() . '/checked', $checklist, [
                     'report-checklist', 'checklist-information'
                 ]);
+            }
+            if(!$noticeAlreadyThere) {
```
slizzio: To avoid confusion: consider putting your if condition directly inline with an explanation comment:
if (!$request->getSession()->getFlashBag()->has('notice') {
   // the duplicate notice is because the PDF view action doesn't actually refresh the page and therefor the original 
   // 'saved' notice never gets rendered

PR #1167, opened 2019-03-18
DDPB-2514: Removes CSV stats download link

aiesoftware:
https://opgtransform.atlassian.net/browse/DDPB-2514


aiesoftware:
Good question - it is, yes, but I'd rather do it in step 2 which is to remove the functionality altogether (DDPB-2415). This initial step is to remove the button from admin with as little impact as needed.

gregtyler:
Okay, fair enough!

PR #1171, opened 2019-03-20
DDPB-2465 - Date comparison validators

acsauk-mojd:
Thanks for the help with this one @aiesoftware - I've added the constraint on interim date not being in the future and the behat assertions to confirm this all.


tests/behat/features/prof/03-report/05-deputy-costs.feature
```diff
@@ -159,6 +159,14 @@ Feature: PROF deputy costs
...
     And the step with the following values CAN be submitted:
       | yes_no_profDeputyCostsHasPrevious_1 | yes |
+    And the step with the following values CANNOT be submitted:
+      | deputy_costs_previous_startDate_day   | 1    |
+      | deputy_costs_previous_startDate_month | 1    |
+      | deputy_costs_previous_startDate_year  | 2016 |
+      | deputy_costs_previous_endDate_day     | 1    |
+      | deputy_costs_previous_endDate_month   | 1    |
+      | deputy_costs_previous_endDate_year    | 2015 |
+      | deputy_costs_previous_amount          | 100  |
```
aiesoftware: We should probably test the boundary here by putting the end date 1 day before the start date, and in the happy path in the lines below change them to being on the same day.

acsauk-mojd:
> Consider getting an exact read of 1 year from the start date and comparing against that, to prevent any leap year issues:
> 
> ```
> $oneYearFromStart = clone $startDate;
> $oneYearFromStart->add(new \DateInterval('P1Y'));
> 
> if ($oneYearFromStart < $endDate) {
>     // violation
> }
> ```

Good catch - I actually had a failing test for this exact scenario with between dates in 2015 and 2016.

PR #1173, opened 2019-03-22
DDPB-2465 fix validation errors

acsauk-mojd:
I've had to rejig the validation callback function here as if we had a valid value in amount but an invalid value in date we were still getting a validation error for the amount due to the  `!empty($ic->getAmount()) && !empty($ic->getDate()` check. 

I'd taken the intent of the function to be checking for just one interim cost and associated date to exist hence why I dropped the `foreach` and moved to checking the first index of the `interimCosts` array.

### DIFF
--- a/src/AppBundle/Entity/Report/Traits/ReportProfDeputyCostsTrait.php
+++ b/src/AppBundle/Entity/Report/Traits/ReportProfDeputyCostsTrait.php
@@ -201,11 +201,11 @@ trait ReportProfDeputyCostsTrait
     {
         $ics = $this->getProfDeputyInterimCosts();
 
-        foreach($this->getProfDeputyInterimCosts() as $index => $ic) {
+        foreach($ics as $index => $ic) {
             if ($ics[$index]->getDate() === null && $ics[$index]->getAmount() === null) {
                 return;
             }
-            
+
             if ($ics[$index]->getDate() === null) {
                 $context->buildViolation('profDeputyInterimCost.date.notBlank')->atPath(sprintf('profDeputyInterimCosts[%s].date', $index))->addViolation();
             }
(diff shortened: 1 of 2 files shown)


src/AppBundle/Entity/Report/Traits/ReportProfDeputyCostsTrait.php
```diff
@@ -199,19 +199,17 @@ public function profCostsHowChangedAtLeastOne(ExecutionContextInterface $context
...
     {
-        // skip validation if there is at least one interim with date
-        foreach($this->getProfDeputyInterimCosts() as $ic) {
-            if (!empty($ic->getAmount()) && !empty($ic->getDate())) {
-                return;
-            }
+        $ics = $this->getProfDeputyInterimCosts();
+
+        if ($ics[0]->getDate() === null) {
+            $context->buildViolation('profDeputyInterimCost.date.notBlank')->atPath('profDeputyInterimCosts[0].date')->addViolation();
```
slizzio: what happens if there are no Interim costs? Wont we get an undefined index at 0 ?

acsauk-mojd: Yep good catch - the way the validation checks were running was hiding this during local testing as the `NotBlank()` on the entity was being run first but if we ever change the order of validation this would become a problem so I've added a `!empty()` conditional around them.

src/AppBundle/Entity/Report/Traits/ReportProfDeputyCostsTrait.php
```diff
@@ -199,19 +199,23 @@ public function profCostsHowChangedAtLeastOne(ExecutionContextInterface $context
      */
     public function profCostsInterimAtLeastOne(ExecutionContextInterface $context)
     {
-        // skip validation if there is at least one interim with date
-        foreach($this->getProfDeputyInterimCosts() as $ic) {
-            if (!empty($ic->getAmount()) && !empty($ic->getDate())) {
+        $ics = $this->getProfDeputyInterimCosts();
+
+        foreach($this->getProfDeputyInterimCosts() as $index => $ic) {
```
slizzio: foreach ($ics as $index => $ic)

PR #1176, opened 2019-03-27
DDPB-2392: Send email to OPG when deputy updates contact details

gregtyler:
Builds upon hack day code to send an email to the OPG when a deputy changes the contact details so that they can be updated in casrec.


gregtyler:
I've removed some of the tests as they can't currently pass:

> We can only test emails sent to addresses starting `behat-`, but we can't manually test those. This puts us in a Catch 22 where we can either use manual testing _or_ automated testing. I'm favouring manual testing (which is consistent with other parts of the system).
>
> In the future, we should rewrite the test harness to properly mock _all_ email addresses, which will make these tests possible.

PR #1186, opened 2019-04-02
DDPB-2538: Confirm report unsubmission

gregtyler:
Adds a new confirmation page when unsubmitting a report to ensure that users have selected the right options.

Introduces a new form type which temporarily holds the unsubmission details and provides a yes/no confirmation option.

Most of the changes to existing code are around the `manage` action, which now handles both the original form and the confirmation one.


gregtyler:
Messed up the refs here, recreating as #1187

PR #1187, opened 2019-04-02
 DDPB-2538: Confirm report unsubmission

gregtyler:
Adds a new confirmation page when unsubmitting a report to ensure that users have selected the right options.

Introduces a new form type which temporarily holds the unsubmission details and provides a yes/no confirmation option.

Most of the changes to existing code are around the `manage` action, which now handles both the original form and the confirmation one.

**Rebuild of #1186 because I messed up the refs somehow**

### DIFF
--- a/src/AppBundle/Controller/Admin/Client/ReportController.php
+++ b/src/AppBundle/Controller/Admin/Client/ReportController.php
@@ -9,6 +9,7 @@ use AppBundle\Entity\Report\Report;
 use AppBundle\Exception\DisplayableException;
 use AppBundle\Form\Admin\ReportChecklistType;
 use AppBundle\Form\Admin\UnsubmitReportType;
+use AppBundle\Form\Admin\UnsubmitReportConfirmType;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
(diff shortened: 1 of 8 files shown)


gregtyler:
> Does this work with PRG pattern? Otherwise all good. Will approve once confirmed.

Should be good: [there's a redirect after saving](https://github.com/ministryofjustice/opg-digi-deps-client/pull/1187/files#diff-b0abae63b50b8d7c71c7f2a07b2faa71R120)

src/AppBundle/Controller/Admin/Client/ReportController.php
```diff
@@ -145,9 +145,30 @@ public function manageAction(Request $request, $id)
...
+        // Use URL data
+        $dataFromUrl = $request->get('data') ?: [];
+        isset($dataFromUrl['startDate']) && $form['startDate']->setData(new \DateTime($dataFromUrl['startDate']));
+        isset($dataFromUrl['endDate']) && $form['endDate']->setData(new \DateTime($dataFromUrl['endDate']));
+        isset($dataFromUrl['dueDateChoice']) && $form['dueDateChoice']->setData($dataFromUrl['dueDateChoice']);
+        isset($dataFromUrl['dueDateCustom']) && $form['dueDateCustom']->setData(new \DateTime($dataFromUrl['dueDateCustom']));
+        if (isset($dataFromUrl['unsubmittedSection'])) {
+            $unsubmittedSections = $form['unsubmittedSection']->getData();
+            foreach ($dataFromUrl['unsubmittedSection'] as $index => $name) {
+                $unsubmittedSections[$index]->setPresent(true);
```
acsauk-mojd: Are the indexes of both arrays always going to correspond to the same data?

gregtyler: Almost certainly, but the code/name would be more consistent so I'll use that instead.

gregtyler: @acsauk-mojd Done, no more index relying

PR #1193, opened 2019-04-08
DDPB-2431 Add Pa11y accessibility functionality

digitalali-moj:
See readme file for testing instructions


gregtyler:
Closing PR. This investigation has indicated that Pa11y is a good tool to use, but we need to further investigate how we can automate it and make it useful in our pipeline.

PR #1208, opened 2019-04-24
DDPB 2563 contact details on confirmation

aiesoftware:


src/AppBundle/Controller/ClientController.php
```diff
@@ -56,6 +57,10 @@ public function editAction(Request $request)
                 $this->getMailSender()->send($addressUpdateEmail, ['html']);
             }
 
+            if ($from === 'declaration') {
```
slizzio: might be better to pass the route names here ?

src/AppBundle/Entity/User.php
```diff
@@ -1136,4 +1136,18 @@ public function setTeamNames($teamNames)
         $this->teamNames = $teamNames;
         return $this;
     }
+
+    /**
+     * @return array
+     */
+    public function getAddressNotEmptyParts()
```
slizzio: getMandatoryAddressFields() sounds a bit better

PR #1213, opened 2019-04-30
DDPB-2614: Use mock email sender in non-production environments

gregtyler:
Rather than mocking emails iff they start `behat-`, mock _all_ emails in non-production environments. As well as allowing us to execute tests more safely, this also untangles mock code from production code.

The changes are fairly straightforward:
- Reuse the existing but unused `MailSenderMock`, moving all mock code from `MailSender`
- Use `MailSenderMock` instead of `MailSender` in 'dev' and 'test' configs
- When emulating production mode in non-production environments, use 'test' config rather than 'prod', so emails continue to be mocked
- Add `app_test` entrypoint for use in non-production environments to force 'test' config to be used.


src/AppBundle/Service/Mailer/MailSenderMock.php
```diff
@@ -3,44 +3,70 @@
...
+use Predis\Client as PredisClient;
 use Swift_Attachment;
+use Swift_Mailer;
 use Swift_Message;
-use Symfony\Bundle\FrameworkBundle\Translation\Translator;
-use Symfony\Component\DependencyInjection\Container;
-use Symfony\Component\Validator\Validator;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 class MailSenderMock extends MailSender
```
aiesoftware: I'm not a fan of this class extending the `MailSender`, as it's not a good fit for inheritance. I think we need a `MailSenderInterface` and both `MailSender` and `MailSenderMock` become implementations of that.

src/AppBundle/Service/Mailer/MailSenderMock.php
```diff
@@ -3,44 +3,70 @@
...
+        $this->mailers = [];
+        $this->validator = $validator;
+        $this->redis = $redis;
+    }
+
+    /**
+     * @param string       $name
+     * @param Swift_Mailer $mailer
+     */
+    public function addSwiftMailer($name, Swift_Mailer $mailer)
```
aiesoftware: I believe this same method is already in the parent. That said, if we move to an interface, this probably ought to be a method on that interface. The implementation needs a little further thought; naming it add `addSwiftMailer` and restricting it by typehint doesn't actually let us add any other mailers, which I presume is the point of the `->mailers` property. If we want this option to add more mailers I think it should be along the lines of `public function addMailer($name, MailerInterface $mailer)`, and we create a `SwiftMailer` which implements the `MailerInterface` and wraps the `Swift_Mailer` class.

gregtyler: Have moved this suggestion to a new ticket ([DDPB-2630](https://opgtransform.atlassian.net/browse/DDPB-2630)) to avoid scope creep on this story and because the current implementation, whilst not perfect, is functional.

PR #1214, opened 2019-05-01
DDPB-2615: Give better access to mocked emails in non-production environments

gregtyler:
**Blocked by #1213**

Adds a new page to non-production environments giving access to emails mocked in the environment. This makes manual testing easier, as anyone testing can inspect what emails have been sent and click appropriate links.


app/config/config_dev.yml
```diff
@@ -18,6 +18,15 @@ services:
...
         tags:
             - { name: data_collector, template: "AppBundle:Manage:api-collector", id: "api-collector" }
+    mail_sender:
+        class: AppBundle\Service\Mailer\MailSenderMock
+        arguments: [ "@validator", "@snc_redis.default" ]
+        calls:
+            - [ addSwiftMailer, [ "default", "@mailer.swift_mailers.default"] ]
+    mailer.transport.smtp.default:
+        class: AppBundle\Service\Mailer\Transport\TransportMock
+    mailer.transport.smtp.secure:
```
slizzio: I dont' think this is needed now, all email is secure (inc default) there should be no reference to it in the code

gregtyler: Good catch, done

PR #1215, opened 2019-05-01
Ddpb 2610 extend document expiry

slizzio:


scripts/cron/digideps
```diff
@@ -4,4 +4,4 @@
 # If executed on frontend container, it'll exit immediately
 # It can be launched manually with "--ignore-s3-failures" in order to ignore S3 deletion failure (only in case the S3 files is already deleted)
 #
-*/5 * * * * app php /app/app/console digideps:documents-cleanup --env=prod > /dev/null 2>&1
+#*/5 * * * * app php /app/app/console digideps:documents-cleanup --env=prod > /dev/null 2>&1
```
gregtyler: If we're removing this cron, can we also delete the `DocumentCleanupCommand` class and test?

slizzio: Yes I think so, im not sure how the cron is set up on all the environments but I think you're right, better not have it at all, rather than have it and doing something it shouldnt. Will remove.

slizzio: Removed

PR #1223, opened 2019-05-20
DDPB 2605 cost estimates breakdown changes

aiesoftware:
Essentially the Deputy Costs Estimates section took an input for the estimated amount, and an input for a breakdown of **other** expenses - and the total expected amount was the sum of all inputs. This is a change to prevent totalling the inputs - instead, the breakdown should form part or all of the original estimated amount, i.e it is a further clarification breakdown of how the estimated figure was derived, and **not** a list of additional costs to factor in.

As a result, I removed the `profDeputyEstimateCostsTotal` property from the entity, and updated all behat tests that were asserting against the value. The layout of the summary page is affected, as is the checklist formatting. 

For more info, see:
https://opgtransform.atlassian.net/browse/DDPB-2605
https://opgtransform.atlassian.net/browse/DDPB-2606
https://opgtransform.atlassian.net/browse/DDPB-2608


src/AppBundle/Resources/views/Report/ProfDeputyCostsEstimate/_answers.html.twig
```diff
@@ -47,89 +47,65 @@
...
+                <tr class="behat-region-more-info">
                     <td>
-                        {% if costRow['amount'] is null %}
-                            £{{ 0 | money_format }}
-                        {% else %}
-                            £{{ costRow['amount'] | money_format }}
-                        {% endif %}
+                        {{ 'moreInfo.form.profDeputyCostsEstimateHasMoreInfo.label' | trans }}
                     </td>
-                    {% if not hideEditLink %}
```
slizzio: Can you check this isnt required as i think its pulled in by the review and report PDF templates and has the Edit link in by default ?

PR #1225, opened 2019-05-23
Ddpb 2613 s3 delete

slizzio:
## Purpose
Permanently removes all versions of a file stored in S3, not just the latest version. This is to adhere to our file retention policy.

Fixes [DDPB-2635]

## Approach
The approach taken is to query S3 for all versions of an object then loop through each one and remove it individually. However locally we were using fake S3 to mimic the real S3 service. FakeS3 did not support the commands needed to query an object's version information. So to getr around this I have replaced FakeS3 with localstack (as used by ServeOPG) which provides a much more realistic replication of S3.

## Learning
Use of AWS CLI to run the object query methods directly against S3 server.
Here are some of the commands used:
**Create bucket
aws s3 --endpoint-url=http://localstack:4572 mb  s3://pa-uploads-local

	

Configure virtual addressing style
aws configure set default.s3.addressing_style virtual

test copy command 
 aws s3 cp --endpoint-url=http://localstack:4572 requirements.txt s3://pa-uploads-local/dd_requirements

aws s3 cp --endpoint-url=http://localstack:4572 requirements.txt s3://pa-uploads-local/dd_doc_11_15584428486388

login to aws locally
docker-compose run aws

display bucket
 aws s3api --endpoint-url=http://localstack:4572 list-object-versions --bucket pa-uploads-local**

** Additional:
Our application talks to S3 via the S3 SDK. There are occasions when the S3 ADK intercepts an exception from S3 and fails to rethrow the exception to our application. Instead it returns a list of 'Errors'. These are now being caught and turned into a runtime exception, this is caught and prevents removal of the database entry.
This is testable by removing the obscure methods allowed within the front.tf terraform config file. ListBucketVersions is required in order to call ListObjectVersions method of the SDK. Likewise DeleteObjectVersion is also required by the SDK. Removing this will result in the Errors being returned rather than Exceptions being thrown.

## Checklist
- [x] The product team have tested these changes


.gitignore
```diff
@@ -31,3 +31,4 @@ web/javascripts
 web/js
 web/stylesheets
 tests/behat/behat.yml
+app/cache
```
gregtyler: This shouldn't be necessary any more since the cache is now stored in `var/cache`. If you've got old artifacts in `app/cache` you should just be able to delete them and they won't reappear.

src/AppBundle/Service/File/Storage/S3Storage.php
```diff
@@ -101,14 +102,71 @@ public function delete($key)
...
+             * ListObjectVersions is permitted by ListBucketVersions in IAM.
+             */
+            $objectVersions = $this->s3Client->listObjectVersions([
+                'Bucket' => $this->bucketName,
+                'Prefix' => $key
+            ]);
+
+            $objectResult = [];
+            $objectsToDelete = '';
+            if ($objectVersions instanceof ResultInterface) {
```
aiesoftware: #notablocker It's usually better to test for the negative and return/throw early, to avoid large chunks of nested code. i.e:

```
if (!$objectVersions instanceof ResultInterface) {
    throw new \RuntimeException('Could not remove file: No results returned');
}
```

src/AppBundle/Service/File/Storage/S3Storage.php
```diff
@@ -101,14 +102,71 @@ public function delete($key)
...
+                            Key: ' . $s3Error['Key'] . ', VersionId: ' .
+                            $s3Error['VersionId'] . ', Code: ' . $s3Error['Code'] . ', Message: ' . $s3Error['Message']);
+                    }
+                    $this->log('error', 'Unable to remove key: ' . $objectResult['Errors'] . '  from S3: ' . json_encode($objectResult['Errors']));
+                    throw new \RuntimeException('Could not remove file: ' . $objectResult['Errors']['Message']);
+                }
+
+                $this->log('info', json_encode($results));
+
+                return $results;
```
aiesoftware: #notablocker For readability, chunks of this method could be broken into smaller methods whose names describe the steps you are taking.

.gitignore
slizzio: removed

src/AppBundle/Service/File/Storage/S3Storage.php
slizzio: Revised, created smaller methods

src/AppBundle/Service/File/Storage/S3Storage.php
slizzio: Revised

PR #1226, opened 2019-05-23
DDPB-2556: Add and expose META file

gregtyler:
This was a bit of standard config I missed from first pass: the server needs to expose its version in the file `/meta.json`, so that Jenkins can identify which version to promote from master to preprod/training.

This change adds and exposes that file, so that the final step of the Jenkins process can proceed.


docker/confd/templates/app.conf.tmpl
```diff
@@ -18,6 +18,10 @@ server {
     ssl_certificate         /etc/nginx/certs/app.crt;
     ssl_certificate_key     /etc/nginx/certs/app.key;
 
+    location =/meta.json {
+        alias /var/www/META;
```
hansnqyr: why not write the file as /var/www/meta.json?

gregtyler: Very good point! And it's now written to `/web/meta.json` so we don't need special nginx config 💃

PR #1232, opened 2019-05-30
DDPB-2648: Additional option for deputy declaration

gregtyler:
## Purpose
Report declarations currently assume that the person signing off the report is a named deputy, but this is not always the case and causes confusion. We need to make a clear option in the declaration form for users who are not named deputies.

Fixes [DDPB-2648](https://opgtransform.atlassian.net/browse/DDPB-2648)

## Approach
I added the new wording for the declaration text to both the NDR and non-NDR report pages. I also added a new `not_deputy` option to the `ReportDeclarationType` forms, allowing users to assert that they are submitting on the deputy's behalf.

I also realised that the formatted report was using the wrong translation keys (they must have been copied and pasted from the NDR version), so I fixed that too.

I took the wording from the prototype and comments on [DDPB-2648](https://opgtransform.atlassian.net/browse/DDPB-2648).

## Learning
I'd hoped replace the `declarationTerms0#` translations with one `declarationTerms` translation containing the content. However, since Symfony translations don't support markdown the only way to do this would've been writing full HTML and using the `raw` nunjucks filter. I figured that this wasn't worthwhile since it broke our practices and left the declaration fields vulnerable to misuse.

## Checklist
  - Uninstalled `gulp-watch` which had a vulnerability in `braces`, but wasn't being used
  - Upgraded Symfony to 3.4.28 and Twig to 1.41.0 to avoid known vulnerabilities


Gulpfile.js
```diff
@@ -172,6 +172,6 @@ gulp.task('watch', gulp.series(setDevelopment, function () {
         config.imgSrc + '/**/*',
         config.jsSrc + '/**/*.js',
         config.jsSrc + '/*.js'],
-        { interval: 1000 },
+        { interval: 1000, usePolling: true },
```
slizzio: Not sure what this is? Is this wanted?

gregtyler: It lets you run Gulp in the background and automatically rebuild whenever you change a JS or CSS file. It's potentially useful when doing a bunch of small frontend changes, but I don't imagine it'd be used often. It doesn't cost us anything to have the option.

src/AppBundle/Resources/views/Report/Report/declaration.html.twig
```diff
@@ -48,7 +48,15 @@
         <div class="text">
             <p>{{ 'page.declarationTerms01' | trans }}</p>
             <p>{{ 'page.declarationTerms02' | trans }}</p>
-            <p>{{ 'page.declarationTerms03' | trans }}</p>
+            {% if report.isLayReport %}
```
aiesoftware: I wonder if this should be driven by the `user.role_name` instead of the report type. If we move towards having different user types accessing the same report (and potentially both being able to reach this declaration page), then we probably want to get away from using a hard coded report type to drive different behaviour based on who is logged in.

PR #1243, opened 2019-06-12
Ddpb 2661 upscale master env

aiesoftware:
## Purpose
Prevent Symfony test env from using mock file system in order to cope with running the test suite across multiple api and frontend instances.

Fixes [DDPB-2661](https://opgtransform.atlassian.net/browse/DDPB-2661)

## Approach
Removing the `session` directive from `config_test.yml` enables the session storage to fall back to the redis storage due to the fact `config_test.yml` inherits from `config_prod.yml`

## Learning
As a result of no longer using the mock session storage, a unit test failed when run across an env with multiple instances. I have decided to remove that unit test file from the app, on the basis that:

* The purpose of the test was to test template logic that determines what content to print depending on the balance status of the report. This feels more suited to Behat (if indeed we want to test this at all), rather than rendering twig content in PHPUnit and using a crawler on the DOM.

* The tests don’t actually prove much - they assert that given a specific report condition, a specific id will exist on a div somewhere on the DOM. This just confirms there is a div with an id, it does not confirm that the contents within that div are correct. The content could be changed, and the test will still pass.

An alternative approach could have been to improve the unit test such that it mocked Symfony's `Form` factory/view functionality, (the error was caused by rendering a form due to CSRF), but the mocks would have been further complicating a unit test which probably needn't exist in the first place.

## Checklist
- [N/A] I have updated documentation (Confluence/GitHub wiki) where relevant
- [N/A] I have added tests to prove my work, and they follow our [best practices](https://github.com/ministryofjustice/opg-digi-deps-client/wiki/Testing-best-practices)
- [N/A] There are no new frontend linting errors (`docker-compose run --rm npm run lint`)
- [N/A] There are no NPM security issues (`docker-compose run --rm npm audit`)
- [N/A] There are no Composer security issues (`docker-compose run frontend php app/console security:check`)


tests/phpunit/Resources/views/Report/BalanceTest.php
```diff
@@ -1,264 +0,0 @@
-<?php
```
slizzio: Why does this test need removing?

aiesoftware: I mentioned it in the PR description.

PR #1246, opened 2019-06-17
DDPB-2679: Hide Behat controller in production

gregtyler:
## Purpose
The Behat controllers are used for debugging and test automation in non-production environments. However, due to a misconfiguration, they are currently available in production too.

Fixes [DDPB-2679](https://opgtransform.atlassian.net/browse/DDPB-2679)

## Approach
The controller now properly throws an exception rather than returning it (the parent function was ignoring it).

I also removed functionality which was inspecting and resetting log files. Now that log messages are put to stderr (and therefore available in CloudWatch), this functionality isn't necessary (and doesn't work).

## Learning
I considered trying to maintain the old behaviour by reading from stderr, but that's not possible in PHP. However, whilst the old behaviour is potentially convenient, I also think it's misleading and insecure.

## Checklist
- [ ] I have successfully built my branch to a feature environment

### DIFF
--- a/tests/behat/bootstrap/LinksTrait.php
+++ b/tests/behat/bootstrap/LinksTrait.php
@@ -27,9 +27,6 @@ trait LinksTrait
         }
     }
 
-    /**
-     * @Given I visit the behat link :link
-     */
     public function visitBehatLink($link)
     {
         $secret = md5('behat-dd-' . $this->getSymfonyParam('secret'));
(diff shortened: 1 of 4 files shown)


tests/behat/bootstrap/LinksTrait.php
```diff
@@ -27,19 +27,13 @@ public function linkWithTextContains($text, $expectedLink)
         }
     }
 
-    /**
-     * @Given I visit the behat link :link
-     */
     public function visitBehatLink($link)
```
slizzio: I often use these comments to determine the actual step to write? Any reason for removing them?

gregtyler: In the case of those specific two, they’re basically private functions so shouldn’t be used in tests directly. Instead we should be calling the test steps which utilise them (e.g. “I open the :regexpr link from the email” ultimately uses `visitBehatLink('get-last-email')`.

PR #1253, opened 2019-06-24
 DDPB-2629: Upgrade PHP to 7.3

gregtyler:
## Purpose
PHP 5.x has reached EOL, so we need to upgrade to PHP 7.

Fixes [DDPB-2629](https://opgtransform.atlassian.net/browse/DDPB-2629)

## Approach
The fundamental changes here are pretty straightforward: use PHP7 base images in the Dockerfile and simplify the extension configuration now that we can use the latest versions.

A few other changes:
- Upgraded `jms/serializer-bundle` to version 3.x since I was seeing _massive_ performance degredations with version 2.x (an order of ~5x)
- Removed `zend_extension` configuration strings since these are done as part of `docker-php-ext-install` and `docker-php-ext-enable`
- Replaced `catch (\Exception)` with `catch (\Throwable)` to [respect changes in PHP7](https://www.php.net/manual/en/language.errors.php7.php)
- Fixed a few issues (mostly in tests) where `count()` was being sent null: PHP5 would return 0 but PHP7 throws.
- Directly imported the `macros.html.twig` file in places where it had previously been inferred from a parent template. This is required in Twig 2, which we have upgraded to with the PHP version change.

## Learning
_Any tips and tricks, blog posts or tools which helped you. Plus anything notable you've discovered about DigiDeps_

## Checklist


composer.json
```diff
@@ -8,7 +8,6 @@
         "psr-4": {"DigidepsBehat\\": "tests/behat/bootstrap" }
     },
     "require": {
-        "php": ">=5.5.9",
```
aiesoftware: https://github.com/ministryofjustice/opg-digi-deps-api/pull/560/files#r297599190

src/AppBundle/Resources/views/Admin/Client/Report/manageConfirm.html.twig
```diff
@@ -1,5 +1,7 @@
 {% extends 'AppBundle:Layouts:application.html.twig' %}
 
+{% import 'AppBundle:Macros:macros.html.twig' as macros %}
```
aiesoftware: I'm curious about these imports as I presume they are already imported from `application.html.twig`?

gregtyler: They are, and in Twig 1.x the import would have been passed down. In Twig 2.x however, imports are done at file level so you need to specifically import the macros file wherever it's used.

This means that:
- Import sources are clearer
- Twig files can be reused more confidently

It's probably also worth noting that, whilst I had to do this in a bunch of places, _most_ files were already importing the macros file even though they didn't technically need to.

aiesoftware: Good to know, thanks

PR #1258, opened 2019-07-01
Ddpb 2669 saaa gifts

slizzio:
## Purpose
Add a Save and Add another to gifts, replacing the need to visit an unnecessary page in journey

Fixes [DDPB-2669]

## Approach
Used existing macros to add the buttons and amended the redirects depending on which button is pressed.

## Checklist
- [ ] There are no new frontend linting errors (`docker-compose run --rm npm run lint`)
- [ ] There are no NPM security issues (`docker-compose run --rm npm audit`)


src/AppBundle/Resources/views/Report/Gift/_form.html.twig
```diff
@@ -10,6 +12,12 @@
         {{ form_select(form.bankAccountId, 'form.bankAccount') }}
     {% endif %}
 
-    {{ macros.saveAndContinueButton(form.save) }}
+    {{ macros.saveAndContinueButton(form.saveAndContinue) }}
+
+    {% if form.saveAndAddAnother is defined %}
+        <span class="push-half--sides text--center">or</span>
```
gregtyler: This is using old GOV.UK Elements classes. I believe the classes should be `govuk-!-margin-left-3 govuk-!-margin-right-3`. I don't think `text--center` would do anything useful in this case.

gregtyler: Sorry, correction: This is using some custom DigiDeps classes which are now part of the Design System, so we should use the Design System versions instead. `govuk-!-margin-left-3 govuk-!-margin-right-3` should do it.

slizzio: will fix. Ta

slizzio: Fixed, applied gov uk style classes to macros and removed span altogether

PR #1278, opened 2019-07-09
DDPB-2750: Only add details banner class for non-lay deputies

gregtyler:
## Purpose
DDPB-2673 introduced a slight regression to the helpline box for lay deputies. This is because it is being given an additional CSS class which aligns and sizes it as for PA and Professional deputies. It's unlikely this would be noticed by any users, but in certain situations it pushes header down just a bit too far.

Fixes [DDPB-2750](https://opgtransform.atlassian.net/browse/DDPB-2750)

## Approach
I only included the class for non-lay reports. I used the existing if statement, but used a temporary variable to stop us calling it three times.

## Checklist
- [x] The product team have tested these changes
  -  N/A

### DIFF
--- a/src/AppBundle/Resources/views/Org/ClientProfile/_detailsBanner.html.twig
+++ b/src/AppBundle/Resources/views/Org/ClientProfile/_detailsBanner.html.twig
@@ -1,5 +1,11 @@
 {% if app.user and (app.user.isDeputyProf() or app.user.isDeputyPa()) and report is defined and report is not null
    and report.client is not null %}
+   {% set showDetailsBanner = true %}
+{% else %}
+    {% set showDetailsBanner = false %}
+{% endif %}
+
+{% if showDetailsBanner %}
 <div class="opg-details-banner">
     <div class="opg-details-banner__content behat-region-client-details-banner">
         <div>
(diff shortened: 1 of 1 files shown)


src/AppBundle/Resources/views/Org/ClientProfile/_detailsBanner.html.twig
```diff
@@ -29,7 +35,6 @@
             {% endif %}
         </h3>
     </div>
-{% if app.user and (app.user.isDeputyProf() or app.user.isDeputyPa()) and report is defined and report is not null
-   and report.client is not null %}
+{% if showDetailsBanner %}
```
aiesoftware: It seems like we don't want this line, and instead the closing div below should be placed before the `{% endif %}` on line 20?

gregtyler: `showDetailsBanner` is controlling a surrounding div. So, if false it looks like:
```twig
<p>Stuff inside</p>
```
and if true:
```twig
<div class="...">
  <p>Stuff inside</p>
</div>
```

So we need to wrap the starting tag and the ending tag of the surrounding div in two different if statements:

```twig
{% if showDetailsBanner %}<div class="...">{% endif %}
  <p>Stuff inside</p>
{% if showDetailsBanner %}</div>{% endif %}
```

aiesoftware: Okay ,that makes sense cheers

PR #1281, opened 2019-07-11
DDPB-2589: Add a remove confirmation page to all summary pages

gregtyler:
## Purpose
In various section summaries, we have the option to edit or remove submitted information. These two buttons are right next to each other, so it's easy (particularly on touch-devices) to accidentally press remove. In most situations this immediately deletes the data without asking for confirmation.

Fixes [DDPB-2589](https://opgtransform.atlassian.net/browse/DDPB-2589)

## Approach
I added guard pages to all sections where we delete data in this way.

I created a standard template which shows the guard page, with a "Yes, remove X" confirmation button, and back/cancel links. Each section passes in some summary and translation data, and the template renders clear and sensible content.

The template pairs with a new `ConfirmDeleteType` form. When this is adequately submitted, the deletion occurs.

I also replaced a few pre-existing deletion guards. They were providing the same functionality as our new form/template, but in a less-consistent way. By using the new method across the board, we now only have one deletion guard template to maintain.

## Learning
This was a good exercise in using a common page template across various different controllers. It seems much cleaner than having duplicate templates to maintain. However, I think certain things can still be improved: translation is quite messy, and I don't like having to pass `$report` into every template just to get the correct details banner.

This change also highlighted how inconsistent our deletion processes are (e.g. with what checks are done, how flashes are created). We could probably do some more work there.

## Checklist


src/AppBundle/Controller/Ndr/AssetController.php
```diff
@@ -338,18 +338,48 @@ public function summaryAction($ndrId)
...
+                ['label' => 'Type', 'value' => 'Property'],
+                ['label' => 'Address', 'value' => implode(', ', $asset->getAddressValidLines())],
+                ['label' => 'Value', 'value' => $asset->getValue(), 'format' => 'money'],
+            ];
+        } else {
+            $summary = [
+                ['label' => 'Type', 'value' => $asset->getTitle()],
+                ['label' => 'Description', 'value' => $asset->getDescription()],
+                ['label' => 'Value', 'value' => $asset->getValue(), 'format' => 'money'],
+                ['label' => 'Valuation date', 'value' => $asset->getValuationDate(), 'format' => 'date'],
```
slizzio: Can we remove these and just use a <entity>_summary partial or a macro call, this is undoing a lot of our translations

src/AppBundle/Entity/Ndr/BankAccount.php
```diff
@@ -24,12 +26,18 @@ class BankAccount implements BankAccountInterface
...
+     */
     private static $typesNotRequiringSortCode = [
         'postoffice',
         'cfo',
         'other_no_sortcode'
     ];
 
+    /**
+     * @JMS\Exclude
+     */
```
slizzio: Are these exclusions not used?

src/AppBundle/Resources/translations/common.en.yml
```diff
@@ -169,3 +169,10 @@ canYouAnswerNDR: |
   it's complete.
 returnToReport: Return to report overview
 returnToClientProfile: Return to client profile
+
+# Confirm delete page
+deletePage:
+    htmlTitle: Deputy report - remove %subject% | GOV.UK
+    question: Are you sure you want to remove this %subject%?
+    pageTitle: Remove %subject%
+    linkButtonLabel: Yes, remove %subject%
```
slizzio: These could be better off in a dedicated translation file?

src/AppBundle/Entity/Ndr/BankAccount.php
gregtyler: They're needed, otherwise JMS Serializer throws an exception when it tries to access them (which was breaking the page render). This has [already been done on the API](https://github.com/ministryofjustice/opg-digi-deps-api/blob/master/src/AppBundle/Entity/Ndr/BankAccount.php#L20) but I guess has gone unnoticed on client because my code deserialises differently to existing functionality (e.g. I retrieve a bank account directly, not as part of a report).

src/AppBundle/Resources/translations/common.en.yml
gregtyler: 👍 Done

src/AppBundle/Controller/Ndr/AssetController.php
gregtyler: Moved to using translations rather than defining strings inside the controller

PR #1285, opened 2019-07-17
DDPB-2751 ensure email is lowercase to match database. Amended behat …

slizzio:
## Purpose
Ensure password reset works with both upper and lower case input.

Fixes [DDPB-2751](https://opgtransform.atlassian.net/browse/DDPB-2751)

## Approach
strtolower email in User entity when requesting email, in case any database entries are in upper case, and changed api call to reset token with lower case email

## Checklist
** found 78 high severity vulnerabilities in 7310 scanned packages ** Handled in DDPB-2769

### DIFF
--- a/src/AppBundle/Entity/User.php
+++ b/src/AppBundle/Entity/User.php
@@ -406,7 +406,7 @@ class User implements AdvancedUserInterface
      */
     public function getEmail()
     {
-        return $this->email;
+        return strtolower($this->email);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/AppBundle/Entity/User.php
```diff
@@ -406,7 +406,7 @@ public function getUsername()
      */
     public function getEmail()
     {
-        return $this->email;
+        return strtolower($this->email);
```
gregtyler: If we're _always_ treating the email as lowercase in the frontend, does it make sense to do the same in API? Or even to always store it in lowercase when we populate the data?

aiesoftware: I agree - I think we should lowercase all emails at the point they are persisted, and that way they will always travel around the application as lowercase. We then only need to convert any user input used for lookup (password reset, admin search).

slizzio: Yes Ive done this, needs further approval. The setEmail method was always using strtolower() which is why there are no instances of capitals in the database. but for consistency Ive changed the DTOs too

